PR #64, opened 2023-06-06
[DEV-599] - Deploy workflow

kin0992:
#### List of Changes

- Remove workflow that published static website to GitHub pages
- Create CD workflow to publish static website to S3 Bucket

#### Motivation and Context

We want to automate the deploy of new version of static website on the S3 Bucket

Manually, looking on AWS dashboard, checking resources changed their statuses

#### Types of changes

#### Checklist:


.github/workflows/cd.yaml
```diff
@@ -0,0 +1,58 @@
+name: Deploy Static Website to S3
```
datalek, line 1: How about changing the name to `Deploy Website` and the name of the file to `deploy-website`?

```suggestion
name: Deploy Website
```

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  pull_request:
+    branches: '**'
```
datalek, line 5: What do you think if we launch it manually until the preview mode is implemented?

.github/workflows/cd.yaml
```diff
@@ -0,0 +1,58 @@
...
+        run: npm run compile
+
+      - name: Build static website
+        env:
+          GITBOOK_API_KEY: ${{ secrets.GITBOOK_API_KEY }}
+        run: npm run build -w nextjs-website
+
+      - name: Export static website
+        env:
+          GITBOOK_API_KEY: ${{ secrets.GITBOOK_API_KEY }}
```
datalek, line 43: I'm not sure but I think that the `export` command doesn't need the `GITBOOK_API_KEY`

kin0992: You are right, the API key is required during the build phase, where we fetching content from GitBook. I've removed from this step.

.github/workflows/cd.yaml
kin0992: What do you think if we have 2 triggers?
- one manually
- one when the a PR has been merged

PR #87, opened 2023-06-28
Set lock file version

kin0992:
Currently, we are using version `18.x` of Node.JS that comes with `v9` of npm.
As we can see from the [official documentation](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json?v=true#lockfileversion), npm `v9` uses the package-lock.json version 3.

#### Motivation and Context

Some tools change the lockfile because no version is specified. (e.g. Next.JS: when run `npm run dev`, NextJS print a warning
> Lockfile was successfully patched, please run "npm install" to ensure [...]

Setting the lockfile version we should prevent this problems.

#### Types of changes

#### Checklist:


datalek (approved):
🔒

PR #129, opened 2023-07-05
Fix contents of pagoPA guide index page

kin0992:
Fix contents of pagoPA guide index page

Changes made here have been tracked as [checklist in this activity](https://pagopa.atlassian.net/browse/DEV-700).


datalek (approved):
🧢

PR #139, opened 2023-07-06
Fix newsroom component

MarcoPonchia:
#### List of Changes

- Refactor add custom newsroom component
- add disable to LinkButton
- manage coomingsoon newsroom

#### Motivation and Context

Update newsroom component

User can see updated newsroom component in http://localhost:3000/app-io/overview

screenshot-localhost_3000-2023 07 06-15_25_21

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/nextjs-website/src/components/atoms/LinkButton/LinkButton.tsx
+++ b/apps/nextjs-website/src/components/atoms/LinkButton/LinkButton.tsx
@@ -4,13 +4,35 @@ import Link from 'next/link';
 import React from 'react';
 
 type LinkButtonProps = {
-  readonly href: string;
+  readonly href?: string;
   readonly label: string;
   readonly size?: number;
   readonly width?: string;
+  readonly disabled?: boolean;
 };
 
-const LinkButton = ({ label, href, size = 18, width }: LinkButtonProps) => {
+const LinkButton = ({
+  label,
+  href,
+  size = 18,
+  width,
+  disabled = false,
+}: LinkButtonProps) => {
+  const labelComponent = (
+    <Typography
+      fontSize={size}
+      fontWeight={700}
+      color={disabled ? 'lightgray' : 'inherit'}
+      sx={{
+        '&::first-letter': {
+          textTransform: 'capitalize',
+        },
+      }}
+    >
+      {label}
+    </Typography>
+  );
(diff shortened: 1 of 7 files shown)


apps/nextjs-website/src/lib/types/tutorialData.ts
```diff
@@ -1,6 +1,7 @@
 import { Path } from '@/lib/types/path';
 
 export type Tutorial = {
+  readonly coomingsoon?: boolean;
```
kin0992, line 4:
```suggestion
  readonly comingSoon: boolean;
```
what do you think to call it `comingSoon` and to remove the optional? optional to a boolean could be handle as "if present -> true; if not -> false"

apps/nextjs-website/src/components/atoms/LinkButton/LinkButton.tsx
```diff
@@ -4,13 +4,35 @@ import Link from 'next/link';
 import React from 'react';
 
 type LinkButtonProps = {
-  readonly href: string;
+  readonly href?: string;
   readonly label: string;
   readonly size?: number;
   readonly width?: string;
+  readonly disabled?: boolean;
```
kin0992, line 11:
```suggestion
  readonly disabled: boolean;
```
same here, as what said with another comment: if you like the idea to not have the `?` on a boolean, then I'd change this as well

apps/nextjs-website/src/lib/types/tutorialData.ts
MarcoPonchia: I rename the attribute as suggested :+1:  But i think is better mark attribute as optional bacause is only temporary flag as coomingSoon content and i think it will be removed in feature.

apps/nextjs-website/src/components/atoms/LinkButton/LinkButton.tsx
MarcoPonchia: Also here i prefer to use optional but for different reason. In this case i have marked `disabled` attribue as optional in props but set to `false` in react component.  So disabled will be always a boolean with default value inside  `LinkButton`  component but can be omitted when used.

kin0992 (approved):

PR #141, opened 2023-07-06
[DEV-655]: add matomo analytics

jessc93:
#### List of Changes

Add Matomo Analytics

#### Motivation and Context

As Admin, I want to collect user navigation data so we know which pages are the most visited

#### Types of changes

#### Checklist:


apps/nextjs-website/src/pages/_document.tsx
```diff
@@ -1,10 +1,32 @@
 import { Html, Head, Main, NextScript } from 'next/document';
 
 export default function Document() {
+  const matomo = `
```
kin0992, line 4: We'd like to make sure matomo is running only on certain condition (e.g. when a variable has a defined value).
Could you update the PR handling that?

The ratio is: if we are on production environment, then we want to track with matomo; otherwise we don't need to track anything with matomo

apps/nextjs-website/src/pages/_document.tsx
```diff
@@ -1,10 +1,35 @@
...
+    g.src = "//cdn.matomo.cloud/pagopa.matomo.cloud/matomo.js";
+    s.parentNode.insertBefore(g, s);
+  })();
+  `;
+
   return (
     <Html lang='en'>
       <Head>
         <meta name='robots' content='noindex,nofollow' />
+        {environment === 'production' && (
```
datalek, line 27:
```suggestion
        {environment === 'prod' && (
```

datalek (changes requested):

datalek (approved):

PR #218, opened 2023-07-31
[DEV-737] - Guide api url

tommaso1:
#### List of Changes

Now `/send/api` supports an additional param called `spec`, if provided the page will try to open the spec corresponding to the name provided in the url, if not found opens the first one as before.

For example opening `http://localhost:3000/send/api?spec=API%20B2B%20avanzamento%20notifiche` will result in 
`API B2B avanzamento notifiche` selected as default

#### Motivation and Context

This can be used to provide a direct link to one specific api if a page contains more than one

opening `http://localhost:3000/send/api?spec=API%20B2B%20avanzamento%20notifiche` will result in 
`API B2B avanzamento notifiche` selected as default  
opening `http://localhost:3000/send/api` will result in the default page being opened

image

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation.


package.json
```diff
@@ -14,6 +14,6 @@
   },
   "devDependencies": {
     "husky": "^8.0.3",
-    "turbo": "^1.9.6"
+    "turbo": "^1.10.12"
```
kin0992, line 17: Why do you need to to update `turbo`?
Is it required to let this functionality work?

tommaso1: it does not work on arm mac os otherwise

kin0992: @tommaso1 Ok, maybe in future PR's I'd add this kind of information within the PR's template to share it with the whole team.

Anyway, IMHO I see two options:
- Create a separate PR with the bump of turbo, explaining why it's needed
- Bump turbo here and it would be nice if you add something more about this into the PR's template

Please note that if you follow Option 1, then [you must update `package-lock.json` file into this PR](https://github.com/pagopa/developer-portal/actions/runs/5724798142/job/15512020098?pr=218#step:4:5)

kin0992:
@tommaso1 I'd suggest you to take a look at this [comment on Jira about the behaviour we want](https://pagopa.atlassian.net/browse/DEV-737?focusedCommentId=90921) for this page.

The idea is that the refresh should still show the very same specification I was watching, but the implementation doesn't reflect that.

tommaso1:
Now it is pushing the param to the url, but because of the behaviour of Elements, this won't work on subpaths

kin0992:
@tommaso1 there is a [check that is failing](https://github.com/pagopa/developer-portal/actions/runs/5807867586/job/15743519974?pr=218#step:4:5). Please take care of the failing workflow and then, if you feel the activity is ready, re-submit the review.

apps/nextjs-website/src/components/molecules/ApiSection/ApiSection.tsx
```diff
@@ -55,8 +56,30 @@ const ApiSection = ({
...
+
   const handleChange = (event: SelectChangeEvent) => {
     setSelectedItemURL(event.target.value);
+
+    const spec = specURLs.find((item) => item?.url === event.target.value);
+
+    if (specURLsName) {
+      // update the url with the spec query param
+      const currentUrl = router.asPath.split('?')[0];
+      router.replace(`${currentUrl}?spec=${spec?.name}`);
```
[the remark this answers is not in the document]
tommaso1: next js router handles it, by automatically encoding the url, the url in the description, `http://localhost:3000/send/api?spec=API%20B2B%20avanzamento%20notifiche` was created just by selecting it on the menu. 

Btw it does not hurt having it there, especially if the url is not created by selecting the menu but if it comes from other sources

kin0992:
@tommaso1 latest changes doesn't work as expected. Going to `/send/api` and then selecting the other OpenAPI from the select list, will generate the error.

tommaso1:
It appears that the suggestion 
```    
if (specURLsName && spec) {
    // update the url with the spec query param
    router.replace(`${product.subpaths.api?.path}?spec=${spec.name}`);
}
```

was the issue, reverting it to the working version

kin0992 (approved):
👍

PR #249, opened 2023-09-05
poc stoplight element loader

tommaso1:
#### List of Changes

POC of a mechanism to show a loader, on api page while stoplight api component loads data. This is just a "loading" div placed below the spotlight component, so that is visible only when the component is not. This way even if the component does not have hooks on the loading we can show a loader

#### Motivation and Context

Stoplight does not contains a loader for their api page, even looking at their example they use an external loader before the component is loaded

#### Types of changes

#### Checklist:


MarcoPonchia:
this PR is outdated after refactor to app router and has been replaced by #314

PR #257, opened 2023-09-14
[DEV-738] - Add different links to Send API using 'spec' query param

marcobottaro:
#### List of Changes

Add different links to Send API using 'spec' query param

#### Motivation and Context

The same link was appearing twice. With 'spec' query param is now possible to distinguish between the two API

Manually

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/nextjs-website/src/_contents/send/apiPath.ts
+++ b/apps/nextjs-website/src/_contents/send/apiPath.ts
@@ -5,12 +5,16 @@ export const sendApiPath: Path = {
   path: '/send/api',
 };
 
+const sendB2BApiPathName = 'API B2B per le Pubbliche Amministrazioni';
 export const sendB2BApiPath: Path = {
-  name: 'B2B API',
-  path: `${sendApiPath.path}?spec=API%20B2B%20per%20le%20Pubbliche%20Amministrazioni`,
+  name: sendB2BApiPathName,
+  path: `${sendApiPath.path}?spec=${encodeURIComponent(sendB2BApiPathName)}`,
 };
 
+const sendNotificationsApiPathName = 'API B2B avanzamento notifiche';
 export const sendNotificationsApiPath: Path = {
-  name: 'Notification API',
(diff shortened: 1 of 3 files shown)


apps/nextjs-website/src/_contents/send/apiPath.ts
```diff
@@ -4,3 +4,13 @@ export const sendApiPath: Path = {
   name: 'API',
   path: '/send/api',
 };
+
+export const sendB2BApiPath: Path = {
+  name: 'B2B API',
```
kin0992, line 9: Avoid duplication of names with https://github.com/pagopa/developer-portal/blob/d59b4097d13e8cd92dfb666f9d58e5a18ce89c5d/apps/nextjs-website/src/_contents/send/api.ts#L12

apps/nextjs-website/src/_contents/send/apiPath.ts
```diff
@@ -4,3 +4,13 @@ export const sendApiPath: Path = {
...
+  path: `${sendApiPath.path}?spec=API%20B2B%20per%20le%20Pubbliche%20Amministrazioni`,
```
kin0992, line 10: Once you have the name, you can use it and put it in the path

apps/nextjs-website/src/_contents/send/apiPath.ts
kin0992: @marcobottaro it looks like you are not avoiding duplication with the file I linked in the comment.

The idea is to use the same variable you just created also [here](https://github.com/pagopa/developer-portal/blob/d59b4097d13e8cd92dfb666f9d58e5a18ce89c5d/apps/nextjs-website/src/_contents/send/api.ts#L12), so you are preventing any duplication errors. Moreover, the React component uses the names inside the `api.ts` file linked to select the proper OpenAPI specification, so if the names are different the page could show an error.

kin0992 (approved):

PR #304, opened 2023-10-04
[DEV-975] - Use shell syntax to read environment variables

kin0992:
> The best practice to avoid code injection vulnerabilities in GitHub workflows is to set the untrusted input value of the expression to an intermediate environment variable and then use the environment variable using the native syntax of the shell/script interpreter (that is, not ${{ env.VAR }}).

According to GitHub, within `run` or `scripts` in GitHub action, [it is more safe to use the shell syntax to get environment variable](https://github.blog/2023-08-09-four-tips-to-keep-your-github-actions-workflows-secure/#1-dont-use-syntax-in-the-run-section-to-avoid-unexpected-substitution-behavior).

This should fix the security issue https://github.com/pagopa/developer-portal/security/code-scanning/1

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/actions/branch-up-to-date/action.yaml
+++ b/.github/actions/branch-up-to-date/action.yaml
@@ -11,7 +11,7 @@ runs:
         TARGET_BRANCH: origin/${{ github.base_ref }}
         HEAD_BRANCH: origin/${{ github.head_ref }}
       run: |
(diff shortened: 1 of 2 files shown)


datalek (approved):
✅

PR #310, opened 2023-10-05
[DEV-746] Generate and add metadata tags on pages

marcobottaro:
#### List of Changes

Generate and add metadata tags on all site pages

#### Motivation and Context

This change is required to properly share developer portal's pages

Manually

#### Types of changes

#### Checklist:


apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx
```diff
@@ -42,6 +43,36 @@ type ProductGuidePageProps = {
...
+  const guideProps = await getGuide(
+    params?.productSlug,
+    params?.productGuidePage ?? ['']
+  );
+
+  const { page } = guideProps;
+
+  const body = page.body;
+  const lines = body.split('\n').filter((line) => line !== '');
+  const title = lines[0].replace('# ', '');
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx:60]: Why you need to do this? Can you add a comment that explains that?

apps/nextjs-website/src/app/[productSlug]/quick-start/page.tsx
```diff
@@ -22,6 +24,28 @@ export type QuickStartGuidePageProps = {
   readonly steps?: ReadonlyArray<Step>;
 } & LayoutProps;
 
+export async function generateMetadata(
+  { params }: ProductParams,
+  parent: ResolvingMetadata
+): Promise<Metadata> {
+  const { shared } = translations;
+  const previousTitle = (await parent).title || shared.siteTitle;
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/quick-start/page.tsx:32]: I often see this line of code. Could be this a function, taking the `ResolvingMetadata` and the `translations` as input and returning the value of the title?
You could also test this function

apps/nextjs-website/src/app/[productSlug]/quick-start/page.tsx
```diff
@@ -22,6 +24,28 @@ export type QuickStartGuidePageProps = {
...
+  const { abstract, path, product } = await getQuickStartGuide(
+    params?.productSlug
+  );
+  const title = `${previousTitle} - ${abstract?.title}`;
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/quick-start/page.tsx:36]: What's happen if the `abstract` is `null` or `undefined`? The title would be something like `Hello - `?

Assume the value of `previousTitle` variable is `Hello`

apps/nextjs-website/src/app/[productSlug]/tutorials/[...productTutorialPage]/page.tsx
```diff
@@ -26,6 +27,34 @@ type ProductTutorialPageProps = {
...
+  params: Params;
+}): Promise<Metadata> {
+  const productSlug = params?.productSlug;
+  const tutorialPath = params?.productTutorialPage?.join('/');
+  const tutorialProps = await getTutorial(productSlug, [tutorialPath]);
+  const { page } = tutorialProps;
+
+  const body = page.body;
+  const lines = body.split('\n').filter((line) => line !== '');
+  const title = lines[0].replace('# ', '');
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/tutorials/[...productTutorialPage]/page.tsx:42]: Same as previous comment: why these lines are needed?
Moreover, it looks like there are the very same line somewhere else, so wouldn't be better to extract those lines in a function and test the function as well?

apps/nextjs-website/src/app/[productSlug]/tutorials/[...productTutorialPage]/page.tsx
```diff
@@ -26,6 +27,34 @@ type ProductTutorialPageProps = {
...
+  return {
+    title,
+    openGraph: {
+      title,
+      url: page.path,
+    },
+    twitter: {
+      site: title,
+      card: 'summary',
+      creator: '@pagopa',
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/tutorials/[...productTutorialPage]/page.tsx:53]: What do you think to avoid duplication? I saw the `@pagopa` into the `twitter` block somewhere else

apps/nextjs-website/src/app/layout.tsx
```diff
@@ -34,6 +35,10 @@ function makeCookieScript(dataDomainScript?: string) {
   `;
 }
 
+export const metadata: Metadata = {
+  metadataBase: new URL('https://developer.pagopa.it'),
```
kin0992, line 39: I honestly don't know what happen adding this line of code. My concern is: what happen when we deploy on dev environment, where the base URL is different?

apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx
```diff
@@ -42,6 +47,30 @@ type ProductGuidePageProps = {
...
+  const title = getTitleFromMarkdown(page.body);
+
+  return {
+    title,
+    openGraph: {
+      title,
+      url: page.path,
+    },
+    twitter: getTwitterMetadata(title),
+  };
```
Comment (kin0992) [inline apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx:71]: Is really necessary to extract the title from the Markdown format?

Wouldn't be easier to do this:
```suggestion
  const {
    page: { path, title },
  } = guideProps;

  return {
    title,
    openGraph: {
      title,
      url: path,
    },
    twitter: getTwitterMetadata(title),
  };
```
the `getGuide` should return the parsed information where you can get the page's title

apps/nextjs-website/src/helpers/metadata.helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export const getTwitterMetadata = (title: string): Twitter => ({
+  title,
+  card: 'summary',
+  site: '@pagopa',
+  creator: '@pagopa',
+});
+
+export const getPreviousTitle = async (
```
kin0992, line 20: why the function should be `async`?

btw, returning a `Promise` you are having issues with the generated `meta` tags value: if you run the application and take a look at the values of the tag you will see something like `content="[object Promise] - L'App IO"`

apps/nextjs-website/src/helpers/metadata.helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+  parent?: ResolvedMetadata
+): Promise<string | AbsoluteTemplateString> => {
```
kin0992, line 22: Are you delegating the caller to handle the returned type?
`AbsoluteTemplateString` is an object, so the caller should handle it. Is this the behaviour we want?

apps/nextjs-website/src/app/page.tsx
```diff
@@ -6,6 +6,25 @@ import Layout from '@/components/organisms/Layout/Layout';
...
+    title: shared.siteTitle,
+    description: homepage.productsShowcaseTitle,
+    openGraph: {
+      title: shared.siteTitle,
+      description: homepage.productsShowcaseTitle,
+      url: '/',
+    },
+    twitter: {
+      card: 'summary_large_image',
+      site: shared.siteTitle,
```
kin0992, line 24: According to [Twitter's documentation about meta tags](https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started), the value of the `twitter.site` property should be "@username for the website used in the card footer."

apps/nextjs-website/src/helpers/metadata.helpers.ts
```diff
@@ -0,0 +1,27 @@
...
+import { AbsoluteTemplateString } from 'next/dist/lib/metadata/types/metadata-types';
+import { Twitter } from 'next/dist/lib/metadata/types/twitter-types';
+
+export const getTitleFromMarkdown = (markdown: string): string => {
+  const lines = markdown.split('\n').filter((line) => line !== '');
+  const titles = lines.filter((line) => line.startsWith('#'));
+  const title = titles[0]?.replace('# ', '').trim() ?? '';
+  return title;
+};
+
```
kin0992, line 12: Take a look if this function is needed or if you can avoid to use it because the title of the pages is parsed by another function (in another comment there is an alternative solution to get the title, but you should check if it's ok for this use case).

If you think this method is still needed, then we should improve it (e.g. you have testes `should return the first title if the markdown has multiple titles` and it's ok, but IMHO you should also tests cases like `should return the first h1 title even it happens after an h2 subtitle` of case where pages do not have title. in this case the function is not working properly)

apps/nextjs-website/jest.config.ts
```diff
@@ -8,6 +8,9 @@ const jestConfig: JestConfigWithTsJest = {
     // See the discussion https://github.com/vercel/next.js/issues/8663
     '^.+\\.(ts|tsx)$': ['ts-jest', { tsconfig: 'tsconfig.jest.json' }],
   },
+  moduleNameMapper: {
```
kin0992, line 11: would you drop a comment why this is needed?

apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx
Comment (MikeAtUqido) [reply apps/nextjs-website/src/app/[productSlug]/guides/[...productGuidePage]/page.tsx:71]: Yeah, you're right. I didn't see the title field on the `getGuide` function, but there is, thanks

apps/nextjs-website/src/app/[productSlug]/guides/page.tsx
```diff
@@ -24,6 +27,26 @@ export type GuidesPageProps = {
...
+
+  return {
+    title,
+    openGraph: {
+      title,
+      description: abstract?.description ?? '',
+      url: path,
+    },
+  };
+};
```
Comment (datalek) [inline apps/nextjs-website/src/app/[productSlug]/guides/page.tsx:48]: No twitter metadata here, is that correct? Also the pages `quick-start`, `overview` and `tutorials` don't have them

apps/nextjs-website/src/helpers/metadata.helpers.ts
datalek, line 1: Consider creating a single `const makeMetadata = (...): Metadata` instead of two different `get*` method.

apps/nextjs-website/src/app/page.tsx
```diff
@@ -6,23 +6,16 @@ import News from '@/components/organisms/News/News';
...
-      description: homepage.productsShowcaseTitle,
-      url: '/',
-    },
-    twitter: {
-      card: 'summary_large_image',
-      site: shared.siteTitle,
-    },
-  };
+    url: '/',
+  });
```
kin0992, line 18: The last time I saw this PR, on the homepage I had the `summary_large_image` as `twitter:card` metatag. Now I'm having that `twitter:card` is equals to `summary`.
Why is this changed?

apps/nextjs-website/src/helpers/metadata.helpers.ts
```diff
@@ -1,27 +1,48 @@
...
 
-export const getTwitterMetadata = (title: string): Twitter => ({
+const getOpenGraphMetadata = (
+  title: string,
+  description?: string,
+  image?: string
+): Metadata['openGraph'] => ({
+  title,
+  type: 'website',
+  locale: 'it_IT',
```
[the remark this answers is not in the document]
MikeAtUqido: At the moment the i18n isn't really done, so I'd wait for that to complete first and the pass the locale as function parameter then. Also because I don't already know how the locale will be provided to the pages.

apps/nextjs-website/src/helpers/metadata.helpers.ts
```diff
@@ -1,27 +1,48 @@
...
+  images: image,
+});
+
+const getTwitterMetadata = (
+  title: string,
+  description?: string,
+  image?: string
+): Metadata['twitter'] => ({
   title,
+  description: description || '',
```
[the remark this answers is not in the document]
MikeAtUqido: Having an empty string or an `undefined` or `null` value is the same in term of meta tags, I'd keep the empty string because some social networks could write the url in the description field and it could not be the best solution to show it. Would you agree?

apps/nextjs-website/src/app/page.tsx
MikeAtUqido: I changed it because we don't really have a `large_image` also for the homepage. So the `summary` would still be a better option

apps/nextjs-website/src/helpers/metadata.helpers.ts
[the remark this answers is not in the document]
kin0992: Yeah, I get it, you could still pass the parameter and setting a default value, then when the i18n will be merged into `main`, you could simply update che caller manipulating the value of the locale and remove the default here in the helper function.

Anyway, it's ok also to wait.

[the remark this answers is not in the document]
MikeAtUqido: No worries, I updated the parameters

kin0992 (approved):

apps/nextjs-website/src/lib/types/makeMetadataFunction.ts
datalek, line 1: What are the benefits of keeping this type here instead of keep them within the file where the [method](https://github.com/pagopa/developer-portal/pull/310/files#diff-e42e2b5e9a618775f9a2c512a6e6d12812c8bc4a7696ba81c5e99d20d71c788bR5) is defined?

kin0992 (approved):

PR #315, opened 2023-10-06
[DEV-977] Fix rendering layout of Embed gitBook block and add custom components…

marcobottaro:
… to render a link with page info and Figma or CodePen's iframes

#### List of Changes

Fix rendering layout of Embed gitBook block and add custom components to render a link with page info and Figma or CodePen's iframes

#### Motivation and Context

The embed block was not working well: the link was broken or empty

Manually

image

#### Types of changes

#### Checklist:


apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
datalek, line 1: Does make sense to add some tests to this component? There are regex, string manipulation and so on

apps/nextjs-website/src/components/organisms/GitBookContent/components/Embed.tsx
```diff
@@ -1,19 +1,63 @@
...
+    case url.includes('youtube.com'):
+      return <EmbedYouTube url={url}>{children}</EmbedYouTube>;
+    case url.includes('checkout.pagopa.it'):
+    case url.includes('docs.italia.it'):
+    case url.includes('io.italia.it'):
+    case url.includes('raw.githubusercontent.com'):
+      return <NotSsrEmbedPageInfo url={url} />;
+    default:
+      return <EmbedLink url={url} />;
+  }
```
datalek, line 60: What about to add some tests to this part? This part can be extracted to a function that given an url return the "type of embed"

apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const [pageTitle, setPageTitle] = useState<string>('');
+  const [siteName, setSiteName] = useState<string>('');
+  const [favicon, setFavicon] = useState<string>('');
+
+  useEffect(() => {
+    async function fetchPageInfo() {
+      // Fetch the URL to get the page's HTML
+      const html = await fetch(url)
+        .then((res) => res.text())
+        .catch((error) => console.error('Error fetching data:', error));
```
datalek, line 41: What about to render a fallback component instead to print an error?

```suggestion
```

datalek (changes requested):

apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
tommaso1: I agree to handle the case of fetch crashing by showing a fallback component (maybe some "this link is broken").

But I'd leave the catch there, because in case of a network error, fetch would throw an exception (all native js api does) and that needs to be catched and handled. 

If the catch stays there, html would be null in case of an error. So in the renderer we could add something like

```
if(!html) {
  return <FallbackComponent/>
}
```

tommaso1 (dismissed):

apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
gunzip: consider that html is null before fetching and on loading. what about using [swr](https://swr.vercel.app/docs/with-nextjs) or react-query for client driven fetches?

apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
marcobottaro: According to what the team decided, I removed the component, so there's no need for tests

apps/nextjs-website/src/components/organisms/GitBookContent/components/Embed.tsx
marcobottaro: Done

apps/nextjs-website/src/components/atoms/EmbedPageInfo/EmbedPageInfo.tsx
marcobottaro: https://github.com/pagopa/developer-portal/pull/315#discussion_r1365477346

packages/gitbook-docs/src/parseContent.ts
```diff
@@ -56,6 +56,7 @@ const schema: ConfigType = {
     swagger,
     code,
     embed,
+    'figma-embed': embed,
```
datalek, line 59: I think this should work properly. You can add a test similar to the one of [embed](https://github.com/pagopa/developer-portal/blob/main/packages/gitbook-docs/src/__tests__/parseContent.test.ts#L613-L645). If it works you can remove the [replaceAll at line 103](https://github.com/pagopa/developer-portal/pull/315/files#diff-6168499bb600557396e98d78cabf1deda878c9ed68e7a18ad25a0ab947a00eb8R103)

```suggestion
    '@figma/embed': embed,
```

datalek (changes requested):

packages/gitbook-docs/src/parseContent.ts
marcobottaro: I added the replaceAll specifically because `@figma/embed` wasn't working

datalek (approved):

MarcoPonchia (approved):

PR #317, opened 2023-10-12
[DEV-852] Menu item should be active if subpath is included in path

MikeAtUqido:
#### List of Changes

* The product header menu item should be active even when the subpath is included in the path

#### Motivation and Context

The menu item lost the active prop when navigating in a subpath

Manually

https://github.com/pagopa/developer-portal/assets/146933303/e79daa25-6fd7-41ac-9016-2306d057dcb6

#### Types of changes

#### Checklist:


apps/nextjs-website/src/helpers/productHeader.helper.ts
```diff
@@ -13,7 +13,7 @@ export function productToMenuItems(
       return {
         label: subpath.name,
         href: subpath.path,
-        active: path === subpath.path,
+        active: path.includes(subpath.path),
```
kin0992, line 16: Could you add a unit test that catches this issue?

If we want to be more sure to not having troubles with paths I'd use the `startsWith` method, what do you think?
The corner case is:
`path` is `/a/b/c-d/e` and `subpath.path` is `/b/c-d`

MarcoPonchia (approved):
Everything is ok :+1:  Great that you also add some tests :muscle:

kin0992 (approved):
💯

PR #343, opened 2023-10-23
[DEV-879] Preserve spaces during token processing

marcobottaro:
#### List of Changes

Preserve spaces during token processing

#### Motivation and Context

When converting tables from GitBook as cards, the parser removed all spaces, but sometimes they are needed to break up words inside other tags (strong, link, etc.)

before: 
image

after:
image

#### Types of changes

#### Checklist:

### DIFF
--- a/packages/gitbook-docs/src/__tests__/parseContent.test.ts
+++ b/packages/gitbook-docs/src/__tests__/parseContent.test.ts
@@ -110,92 +110,19 @@ describe('parseContent', () => {
     expect(parseContent('This is a paragraph', config)).toStrictEqual([
       new Markdoc.Tag('Paragraph', {}, ['This is a paragraph']),
     ]);
+  });
 
+  it('should preserve a single space into a td tag nested into other tags', () => {
     expect(
       parseContent(
-        '<strong>Sei un</strong> <a data-footnote-ref href="#user-content-fn-1"><strong>ente aggregatore</strong></a><strong>?</strong>',
+        '<td><strong>before space</strong> <a><strong>after space</strong></a></td>',
         config
       )
     ).toStrictEqual([
-      new Markdoc.Tag('Paragraph', {}, [
-        new Markdoc.Tag('StyledText', { style: 'strong' }, ['Sei un']),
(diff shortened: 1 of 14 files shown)


packages/gitbook-docs/src/markdoc/tokenProcessor.ts
```diff
@@ -29,12 +29,19 @@ const makeHtmlParser = () => {
       result.push(token);
     },
     ontext: (content) => {
-      if (typeof content === 'string' && content.trim().length > 0) {
+      if (typeof content !== 'string') return;
+      if (content.trim().length > 0) {
```
datalek, line 33: What about to replace the `else if` with the following condition?
```suggestion
      if (content.trim().length > 0 || content === ' ') {
```

packages/gitbook-docs/src/__tests__/parseContent.test.ts
```diff
@@ -110,6 +110,94 @@ describe('parseContent', () => {
...
+                ]),
+              ]
+            ),
+            new Markdoc.Tag('StyledText', { style: 'strong' }, ['?']),
+          ]),
+          new Markdoc.Tag('CardItem', {}, []),
+          new Markdoc.Tag('CardItem', {}, []),
+        ]),
+      ]),
+    ]);
```
datalek, line 200: What about to simplify as following? It should cover all the cases we need

```suggestion
  });

  it('should preserve a single space into a td tag nested into other tags', () => {
    expect(
      parseContent(
        '<td><strong>before space</strong> <a><strong>after space</strong></a></td>',
        config
      )
    ).toStrictEqual([
      new Markdoc.Tag('StyledText', { style: 'strong' }, ['before space']),
      ' ',
      new Markdoc.Tag('Link', {}, [
        new Markdoc.Tag('StyledText', { style: 'strong' }, ['after space']),
      ]),
    ]);
```

datalek (commented):
Good job! I've just suggested some simplification

kin0992 (approved):
👍

PR #344, opened 2023-10-23
[DEV-1010] - Update user pool's password policy

kin0992:
According to the [OWASP Authentication Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls), a password is weak if it has less than `8` chars.
It should also allow usage of all characters including unicode and whitespace to be considered as a `strong` password.

#### List of Changes

- Password must contain lowercase character
- Password must contain uppercase character
- Password must contain at least a number
- Password must contain at least a symbol

#### Motivation and Context

We want to make sure users' passwords are considered `strong`.

For further details about Cognito user pool's password policy, [check the documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-policies.html).

Manually. The policy has been updated, without the need of deleting the Cognito User Pool.
It does not force users with a _legacy_ password to change it.

#### Types of changes

#### Checklist:

### DIFF
--- a/.infrastructure/55_cognito.tf
+++ b/.infrastructure/55_cognito.tf
@@ -23,6 +23,10 @@ resource "aws_cognito_user_pool" "devportal" {
 
   password_policy {
     minimum_length                   = 8
+    require_lowercase                = true
+    require_numbers                  = true
+    require_uppercase                = true
+    require_symbols                  = true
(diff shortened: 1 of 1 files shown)


datalek (approved):
🔑

PR #358, opened 2023-10-31
[DEV-1066] Add webinar and speaker types

marcobottaro:
#### List of Changes

Add webinar and speaker types

#### Types of changes

#### Checklist:


apps/nextjs-website/src/lib/types/webinar.ts
```diff
@@ -0,0 +1,10 @@
+import { Speaker } from '@/lib/types/speaker';
+
+export type Webinar = {
+  readonly description: string;
+  readonly speakers?: readonly Speaker[];
```
kin0992, line 5: Can a webinar have no speakers?

apps/nextjs-website/src/lib/types/webinar.ts
```diff
@@ -0,0 +1,10 @@
...
+  readonly startDateTime?: Date;
```
kin0992, line 6: Also the startDateTime shouldn't be required?

apps/nextjs-website/src/lib/types/webinar.ts
marcobottaro: Yes, I imagined the scenario where the speakers are not yet confirmed (and the date is not sure*), but the devPortal's staff needs to start a campaign about the upcoming webinar

apps/nextjs-website/src/lib/types/webinar.ts
marcobottaro: * for the same reason of the undefined speakers

apps/nextjs-website/src/lib/types/webinar.ts
kin0992: yep, makes sense, and an empty array can let you handle webinars without speakers.
if the `speakers` property can also be `undefined` we should both handle `undefined` values and empty arrays

kin0992 (approved):

PR #371, opened 2023-11-06
[DEV-1088] Fix wrong mailto link

tommaso1:
#### List of Changes

Fix mailto links markdown parsing

#### Motivation and Context

DEV-1088, now the parsing of mailto links is not correct

go to `/pago-pa/guides/sanp/3.5.0/specifiche-attuative-del-nodo-dei-pagamenti-spc/funzionamento-generale/ladesione-alla-piattaforma-pagopa`, click on `pagamenti@assistenza.pagopa.it` and check it behaves like a mailto link

#### Types of changes

#### Checklist:


apps/nextjs-website/src/components/organisms/GitBookContent/components/Link.tsx
```diff
@@ -6,12 +6,19 @@ import { LinkProps } from 'gitbook-docs/markdoc/schema/link';
...
-    href={href}
+    href={extractMailTo(href) ?? href}
     title={title}
     sx={{ textDecoration: 'none' }}
   >
     {children}
   </MUILink>
 );
 
+function extractMailTo(href: string): string | undefined {
```
marcobottaro, line 17: I think we should move this logic into parsing context and add a new test

packages/gitbook-docs/src/markdoc/schema/link.ts
```diff
@@ -18,6 +18,16 @@ export const link: Schema = {
...
     const attrs = node.transformAttributes(config);
+
+    /* eslint-disable functional/no-expression-statements */
+    /* eslint-disable functional/immutable-data */
+    if (attrs.href && attrs.href.includes('mailto:')) {
+      // delete everything before the 'mailto:'
+      attrs.href = attrs.href.replace(/.*mailto:/, 'mailto:');
+    }
+    /* eslint-enable functional/no-expression-statements */
+    /* eslint-enable functional/immutable-data */
```
kin0992, line 29:
```suggestion
    const attrs = pipe(node.transformAttributes(config), (attributes) =>
      pipe(
        attributes.href && attributes.href.includes('mailto:'),
        b.fold(
          () => attributes,
          () => ({
            ...attributes,
            href: attributes.href.replace(/.*mailto:/, 'mailto:'),
          })
        )
      )
    );
```
an alternative way keeping data immutable

tommaso1: That's not working, tests does not pass anymore with that modification, also, i think that 'b' is not the most meaningful name to give to something

packages/gitbook-docs/src/markdoc/schema/link.ts
```diff
@@ -18,6 +19,15 @@ export const link: Schema = {
...
     const attrs = node.transformAttributes(config);
+    /* eslint-disable functional/no-expression-statements */
+    /* eslint-disable functional/immutable-data */
+    if (attrs.href && attrs.href.includes('mailto:')) {
+      // delete everything before the 'mailto:'
+      attrs.href = attrs.href.replace(/.*mailto:/, 'mailto:');
+    }
+    /* eslint-enable functional/no-expression-statements */
+    /* eslint-enable functional/immutable-data */
+
```
datalek, line 30: Instead change the behaviour of `LinkAttr` ([/packages/gitbook-docs/src/markdoc/attributes.ts#L36-L55](https://github.com/pagopa/developer-portal/blob/23e4942/packages/gitbook-docs/src/markdoc/attributes.ts#L36-L55))

```suggestion
```

datalek (changes requested):

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -26,6 +26,11 @@ const convertLink = (link: string): string =>
 // eslint-disable-next-line functional/no-classes
 export class LinkAttr {
   readonly transform = (value: string | null, { variables }: Config) => {
+    if (value && value.includes('mailto:')) {
```
datalek, line 29: I think the `startsWith` suite more the use-case (here the value should be `mailto:...`). Moreover some lines below there is already an if, I think we should extend it adding an `else if`

datalek (changes requested):

packages/gitbook-docs/src/markdoc/attributes.ts
tommaso1: ok for the else if, but i'd leave it to be `includes` because i think itìs more future proof given that the links are not hardcoded but they can be configured externally.

datalek (approved):

kin0992 (approved):
📫

PR #373, opened 2023-11-07
[DPC-91] Update version "Approfondimenti sull’integrazione" SEND

certevol:
#### List of Changes
Change the ref for sync to docs

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/nextjs-website/src/_contents/send/guides.ts
+++ b/apps/nextjs-website/src/_contents/send/guides.ts
@@ -56,6 +56,10 @@ const knowledgeBase: GuideDefinition = {
       dirName: 'LkozJ77TMlb5XupVDO7J',
     },
     {
+      version: 'v2.1',
+      dirName: 'LjuHmRnHyge98hRa4Ao4',
+    },
+    {
       version: 'v1.0',
       dirName: '8hvzBYw259fYwQSqzmf6',
     },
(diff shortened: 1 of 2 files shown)


apps/nextjs-website/src/_contents/send/guides.ts
certevol, line 1: The main version must remain 2.0, for now. So, I just edited this file.

apps/nextjs-website/src/_contents/send/guides.ts
```diff
@@ -55,6 +55,10 @@ const knowledgeBase: GuideDefinition = {
       version: 'v2.0',
       dirName: 'LkozJ77TMlb5XupVDO7J',
     },
+    {
+      version: 'v2.1',
+      dirName: 'LjuHmRnHyge98hRa4Ao4',
+    },
```
datalek, line 61: Why it is not in the position 0? I think the version menu on the guide page maintains this order

datalek (approved):

apps/nextjs-website/src/_contents/send/guides.ts
certevol: because 2.0, temporarily will remain the main version but some users are already using version 2.1. It's dev Vs prod situation.

PR #374, opened 2023-11-07
[DEV-1043] - Add missing sign up parameters to user pool

kin0992:
Add missing sign up parameters to Cognito User Pool.

#### List of Changes

- Add role
- Add company type

#### Motivation and Context

During the sign up, we want to be able also to collect those information.

#### Types of changes

#### Checklist:


.infrastructure/55_cognito.tf
```diff
@@ -169,6 +169,30 @@ resource "aws_cognito_user_pool" "devportal" {
...
 
+  schema {
+    name                     = "role"
+    attribute_data_type      = "String"
+    developer_only_attribute = false
+    mutable                  = true
+    required                 = false
+    string_attribute_constraints {
+      min_length = 1
+      max_length = 256
```
datalek, line 180: The value `max_length` can be changed in the future with an update in place?

kin0992: Unfortunately the [AWS Cognito Doc](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-settings-attributes.html) says it is not possible to update an attributes once you add it to the user pool.

In this case, I'd use the max value possible for the `max_length`, which is `2048`.

.infrastructure/55_cognito.tf
```diff
@@ -169,6 +169,30 @@ resource "aws_cognito_user_pool" "devportal" {
     required                 = false
   }
 
+  schema {
+    name                     = "role"
```
datalek, line 173:
```suggestion
    name                     = "job_role"
```

datalek (approved):

datalek (approved):

PR #380, opened 2023-11-09
[DEV-1056] Add e-mail template for confirmation message

marcobottaro:
#### List of Changes

Add responsive e-mail template for confirmation message using MJML framework. 

#### Motivation and Context

The previous e-mail didn't respect design requirements

Adding `console.log(emailMessage)` after `custom-message-handler.ts:33` in order to print `emailMessage` output in console and running tests with this command:
`npm -w cognito-functions run test -- --watch `

image

#### Types of changes

#### Checklist:


apps/cognito-functions/src/templates/confirmation-message.ts
```diff
@@ -0,0 +1,80 @@
+import { COMPANY_LOGO } from './company-logo';
+
+const TRANSLATIONS: { readonly [k: string]: string } = {
```
marcobottaro, line 3: We should think about e-mail translations and replace this object with something more appropriate

package.json
```diff
@@ -13,7 +13,13 @@
...
   },
   "devDependencies": {
+    "@types/html-minifier": "^4.0.4",
+    "@types/mjml": "^4.7.4",
     "husky": "^8.0.3",
     "turbo": "^1.10.12"
+  },
+  "dependencies": {
+    "html-minifier": "^4.0.0",
+    "mjml": "^4.14.1"
```
kin0992, line 23: you should add these new dependencies only in the `cognito-functions` module

apps/cognito-functions/src/templates/company-logo.ts
kin0992, line 1: I think you should use the [`LogoPagoPACompany ` from MUI-Italia](https://github.com/pagopa/mui-italia/blob/92eeb5c8919bf4ddc2a0ba560a19b569951c36e5/src/assets/LogoPagoPACompany/LogoPagoPACompany.tsx#L30)

kin0992 (changes requested):

package.json
datalek: Check the section [Manage Dependencies](https://github.com/pagopa/developer-portal#manage-dependencies) of the README, there are the examples to show how do it

apps/cognito-functions/src/custom-message-handler.ts
```diff
@@ -1,21 +1,24 @@
...
+export const emailTemplate = (href: string) => {
+  const emailMessage = mjml2html(customMessage(href)).html;
+
+  return minify(emailMessage, {
+    collapseWhitespace: true,
+    minifyCSS: true,
+    caseSensitive: true,
+    removeEmptyAttributes: true,
+  });
+};
```
datalek, line 21: What about move this function into `apps/cognito-functions/src/templates/confirmation-email.ts` and rename it to `makeConfirmationEmail`?

datalek (changes requested):

apps/cognito-functions/src/templates/company-logo.ts
marcobottaro: I don't think we should use React components inside MJML templates. We would be forced to install another [dependency](https://github.com/Faire/mjml-react) and rewrite the whole template with a similar, but different syntax, just to avoid a few rows of code duplication

kin0992: The advantage I see using a library like [dependency](https://github.com/Faire/mjml-react) is that we can re-use components created in the developer-portal and somewhere else (like the icon I linked before). 
If for example we need to add an Hero (and this is just an example popped out my mind) in the email, we need to create it again, using the syntax of mjml while if we could use React components, we could recycle the `Hero` component already developed; and in that way we could improve a lot, in my opinion, the experience of the emails.

Anyway, I think for now it's good the way it is and we can discuss about the usage of React components in email templates later! 🍻

datalek (approved):

kin0992 (approved):

PR #393, opened 2023-11-14
[DEV-1091] Add create auth challenge handler

datalek:
> [!IMPORTANT]  
> The lambda in not yet connected to the cognito trigger. This will be done later once the UI is ready

#### List of Changes

- Add `createAuthChallengeHandler`
- Add terraform lambda resource

#### Motivation and Context

Thee login flow must have the MFA via email. This PR add the logic to generate and send the verification code via email.

Unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/cognito-functions/src/create-auth-challenge-handler.ts
+++ b/apps/cognito-functions/src/create-auth-challenge-handler.ts
@@ -1,4 +1,5 @@
 import * as t from 'io-ts';
+import crypto from 'crypto';
 import { CreateAuthChallengeTriggerEvent } from 'aws-lambda';
 import {
   SES,
@@ -9,6 +10,9 @@ import { pipe } from 'fp-ts/lib/function';
 import * as TE from 'fp-ts/TaskEither';
 import * as T from 'fp-ts/Task';
 import * as E from 'fp-ts/Either';
+
+export const generateVerificationCode = (): string =>
+  Array.from({ length: 6 }, () => crypto.randomInt(0, 9)).join('');
 
 export const emailTemplate = (verificationCode: string) => `
 Il tuo codice di verifica per la login è <strong>${verificationCode}</strong>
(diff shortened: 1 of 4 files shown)


apps/cognito-functions/src/create-auth-challenge-handler.ts
```diff
@@ -0,0 +1,95 @@
...
+  async (
+    event: CreateAuthChallengeTriggerEvent
+  ): Promise<CreateAuthChallengeTriggerEvent> => {
+    const { session } = event.request;
+
+    // only called once after SRP_A and PASSWORD_VERIFIER challenges. Hence
+    // session.length == 2
+    if (session.length === 2) {
+      const { email } = event.request.userAttributes;
+      const verificationCode = env.generateVerificationCode(); // env.crypto.randomBytes(3).toString('hex');
```
datalek, line 61:
```suggestion
      const verificationCode = env.generateVerificationCode();
```

apps/cognito-functions/src/main.ts
```diff
@@ -33,3 +35,22 @@ export const sensEmailHandler = pipe(
...
+      // eslint-disable-next-line functional/no-expression-statements
+      console.log(PR.failure(errors).join('\n'));
+      // eslint-disable-next-line functional/no-throw-statements
+      throw new Error();
+    },
+    (config) =>
+      createAuthChallenge.makeHandler({
+        config,
+        ses: new SES(),
+        generateVerificationCode: () => crypto.randomBytes(3).toString('hex'),
```
kin0992, line 53: I think we should move the `generateVerificationCode` somewhere else and add unit tests to it so we can verify the function will generate OTP in a format we want.
Looking at the tests, we want a 4 digits code, but the `crypto.randomBytes` function returns a 6 chars alphanumeric string.

What do you think?

kin0992 (approved):
🎲

PR #400, opened 2023-11-16
[DEV-1121] - Custom esbuild configuration

kin0992:
Improve esbuild configuration using the `esbuild.config.mjs` file.

#### List of Changes

- Create `esbuild.config.mjs` file
- Add a workaround to allow the bundle generated to work (more details are available [here](https://github.com/mjmlio/mjml/issues/2132#issuecomment-1004713444))

#### Motivation and Context

We had an issue with the `uglify-js` dependency, imported by the `mjml` dependency.

Manually

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/cognito-functions/package.json
+++ b/apps/cognito-functions/package.json
@@ -4,8 +4,10 @@
   "private": true,
   "scripts": {
     "clean": "shx rm -rf dist/ out/",
-    "compile": "npm run clean && tsc",
-    "build": "npm run compile && esbuild dist/main.js --bundle --platform=node --outfile=out/main.js",
+    "precompile": "npm run clean",
+    "compile": "tsc",
+    "prebuild": "npm run compile",
+    "build": "node esbuild.config.mjs",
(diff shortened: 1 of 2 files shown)


datalek (approved):
🏗️

apps/cognito-functions/package.json
```diff
@@ -4,8 +4,10 @@
   "private": true,
   "scripts": {
     "clean": "shx rm -rf dist/ out/",
-    "compile": "npm run clean && tsc",
-    "build": "npm run compile && esbuild dist/main.js --bundle --platform=node --outfile=out/main.js",
+    "precompile": "npm run clean",
+    "compile": "tsc",
+    "prebuild": "npm run compile",
```
gunzip, line 9: what's the need to run `tsc` before `esbuild` during deploys?

kin0992: you are right. we are going to fix it with #405

PR #402, opened 2023-11-16
[DEV-1100] Cookie banner close automatically

jeremygordillo:
#### List of Changes

Replace the `starDateTime` and `endDateTime` props types from Date to string

#### Motivation and Context

These changes solve following error: https://legacy.reactjs.org/docs/error-decoder.html/?invariant=425, that was generate by dates of webinar component.

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/nextjs-website/src/components/atoms/TimeSlot/TimeSlot.tsx
+++ b/apps/nextjs-website/src/components/atoms/TimeSlot/TimeSlot.tsx
@@ -48,14 +48,16 @@ function conditionallyFormattedEndDate(
 }
 
 type TimeSlotProps = {
-  start?: Date;
-  end?: Date;
+  start?: string;
+  end?: string;
 };
 
 const TimeSlot = ({ start, end }: TimeSlotProps) => {
+  const startDate = start ? new Date(start) : undefined;
+  const endDate = end ? new Date(end) : undefined;
   return [
-    start && formattedDateTime(start),
-    conditionallyFormattedEndDate(start, end),
+    start && formattedDateTime(startDate as Date),
+    conditionallyFormattedEndDate(startDate, endDate),
   ]
     .filter(Boolean)
     .join(' - ');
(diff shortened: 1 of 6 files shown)


marcobottaro (approved):

marcobottaro (approved):

marcobottaro (approved):

apps/nextjs-website/src/lib/types/webinar.ts
```diff
@@ -5,8 +5,8 @@ export type Webinar = {
   readonly description: string;
   readonly html?: string;
   readonly speakers: readonly Speaker[];
-  readonly startDateTime?: Date;
-  readonly endDateTime?: Date;
+  readonly startDateTime?: string;
```
datalek, line 8: Write a comment about why this should be a string.

apps/nextjs-website/src/components/atoms/TimeSlot/TimeSlot.tsx
datalek, line 1: I think there is too untested stuff here. Add a ticket to test this component

datalek (approved):

apps/nextjs-website/src/lib/types/webinar.ts
jeremygordillo: @datalek In the issue description you can find the reason for these changes.
Do you want me to add the same comment in the code too?

apps/nextjs-website/src/components/atoms/TimeSlot/TimeSlot.tsx
marcobottaro: I added this task in the backlog: https://pagopa.atlassian.net/browse/DEV-1138

apps/nextjs-website/src/lib/types/webinar.ts
marcobottaro: @jeremygordillo here you should only explain (the repo is public, but the board is not) why you converted a Date to a string in the Webinar type.
I suppose that on the one hand, you expect that the API will send a string, and on the other hand, converting to date only in the component you resolve the problem of the non-corresponding dates, am I right?

jeremygordillo: Yes, you're right. I have improved the PR description with the explanation of these changes.

kin0992 (approved):

PR #431, opened 2023-11-21
[DEV-1152] Disable submit button on form submission

jeremygordillo:
#### List of Changes

I added submission state to enable or disable submit buttons

#### Motivation and Context

These changes are necessary because the user can send data multiple times without waiting for a response, so to improve the UX, i added a submission state where the buttons are disabled until the server responds.

#### Types of changes

#### Checklist:


marcobottaro (approved):

apps/nextjs-website/src/components/organisms/Auth/ConfirmLogin.tsx
```diff
@@ -28,10 +28,16 @@ const ConfirmLogin = ({ onBackStep, onConfirmLogin }: confirmLoginProps) => {
...
 
-  const onconfirmLoginHandler = useCallback(() => {
-    onConfirmLogin(code).catch((e) => setError(e.message));
+  const onConfirmLoginHandler = useCallback(() => {
+    setSubmitting(true);
+    onConfirmLogin(code)
+      .catch((e) => setError(e.message))
+      .finally(() => {
+        setSubmitting(false);
+      });
```
marcobottaro, line 40: If you set `submitting` to `true` there is a time - when the user has been loaded, but the redirect to the home page has not yet been completed - in which you can click the submit button again. 

```suggestion
    onConfirmLogin(code).catch((e) => setError(e.message));
```

marcobottaro (changes requested):

apps/nextjs-website/src/components/organisms/Auth/ConfirmLogin.tsx
jeremygordillo: I removed the finally step and moved the `setSubmitting(false)` call inside the catch step because the only case in which we need to reset the submission state is when the onConfirm callback throws an error otherwise, on success, the user will be redirected to homepage.

MarcoPonchia (approved):

marcobottaro (approved):

PR #435, opened 2023-11-22
[DEV-1162] remove {% p %} on table children

tommaso1:
#### List of Changes

Remove `{% p %}` from table children text because markdoc parses it incorrectly

go to /pago-pa/guides/saci/3.1.0/specifiche-attuative-dei-codici-identificativi-di-versamento-riversamento-e-rendicontazione/flusso-di-rendicontazione, check there are no more `{% p %}` around

Untitled

#### Types of changes

#### Checklist:


datalek (commented):
Add a test that covers this use case.

I think the problem is caused by [the old workaround](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L96-L102) that replaces some `html` tags to Markdoc tags (`{% tag %}`). One of the replaced tags is the Paragraph (from `<p>` to `{% p %}`). 

I'd remove the [workaround from the Paragraph tag](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L100), and leave the system to manage it as html token (check the [comments on these lines](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L105-L110))

datalek (changes requested):
I would try what is described in this [comment](https://github.com/pagopa/developer-portal/pull/435#pullrequestreview-1744642080)

MarcoPonchia:
> Add a test that covers this use case.
> 
> I think the problem is caused by [the old workaround](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L96-L102) that replaces some `html` tags to Markdoc tags (`{% tag %}`). One of the replaced tags is the Paragraph (from `<p>` to `{% p %}`).
> 
> I'd remove the [workaround from the Paragraph tag](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L100), and leave the system to manage it as html token (check the [comments on these lines](https://github.com/pagopa/developer-portal/blob/a77bf31/packages/gitbook-docs/src/parseContent.ts#L105-L110))

I close this outdated PR to solve this issue with paragraph following this suggestion above in a new dedicated PR to the refactor

PR #498, opened 2023-12-07
[DPC-119] Update version SACI & SANP

certevol:
Depends on #514 

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #514, opened 2023-12-13
[DEV-1229] Manage missing closing file tag in markdown files

jeremygordillo:
#### List of Changes

Added regex to find "file" tags without closing tag and for each match adds it

#### Motivation and Context

To fix parsing when `file` tag has not a closing tag

[Caricamento massivo iban](http://localhost:3000/pago-pa/guides/manuale-bo-ec/v1.0/manuale-operativo-back-office-pagopa-ente-creditore/funzionalita/gestione-iban/caricamento-massivo-iban)

[Modelli base](http://localhost:3000/pago-pa/guides/avviso-pagamento/3.2.0/allegato-1/modelli-base)

image

#### Types of changes

#### Checklist:


marcobottaro:
[/pago-pa/guides/manuale-bo-ec/v1.0/manuale-operativo-back-office-pagopa-ente-creditore/funzionalita/gestione-iban/caricamento-massivo-iban](https://github.com/pagopa/developer-portal/pull/498#issuecomment-1849702173)

marcobottaro (approved):

datalek (changes requested):
It's not clear when the old implementation fails. I would suggest adding a test case that fails with the old implementation but doesn't with the new one

jeremygordillo:
> It's not clear when the old implementation fails. I would suggest adding a test case that fails with the old implementation but doesn't with the new one

@datalek The error was generated by this specific case:

```md
{% tabs %}
{% tab title="Tab 1" %}
Tab 1 Content

{% file src="path/to/file" %}
{% endtab %}

{% tab title="Tab 2" %}
Tab 2 Content

{% file src="path/to/file" %}
{% endtab %}
{% endtabs %}
```

When we have a file without closing tag inside a tab, the parser performs incorrect parsing. So, to avoid problems with this case, I created a regex to add the closing tag when it is missing.

packages/gitbook-docs/src/__tests__/parseContent.test.ts
```diff
@@ -560,6 +561,30 @@ describe('parseContent', () => {
...
+        new Markdoc.Tag('Paragraph', {}, [
+          new Markdoc.Tag('Paragraph', {}, ['Tab 1 Content']),
+          new Markdoc.Tag('File', fileAttrs),
+        ]),
+        new Markdoc.Tag('Paragraph', {}, [
+          new Markdoc.Tag('Paragraph', {}, ['Tab 2 Content']),
+          new Markdoc.Tag('File', fileAttrs),
+        ]),
+      ]),
+    ]);
```
datalek, line 587: It seems that the test does not break with the old implementation; I would expect it to break with the old implementation

datalek (approved):
LGTM

packages/gitbook-docs/src/__tests__/parseContent.test.ts
datalek: My mistake, I didn't take the regex into account.

PR #559, opened 2024-01-11
[DEV-1234] woff2 font format is not loaded

jeremygordillo:
#### List of Changes

Updated lambda checks to prevent `.woff2` font requests from being modified

#### Motivation and Context

This change is required to load fonts with `.woff2` format

#### Types of changes

#### Checklist:

### DIFF
--- a/.changeset/eighty-toys-care.md
+++ b/.changeset/eighty-toys-care.md
@@ -1,0 +1,5 @@
+---
+"cloudfront-functions": patch
+---
+
+handle woff2 font request
--- a/.changeset/eighty-toys-care.md
+++ b/.changeset/eighty-toys-care.md
@@ -2,4 +2,4 @@
 "cloudfront-functions": patch
 ---
 
-handle woff2 font request
+Handle woff2 font request
--- a/apps/cloudfront-functions/src/viewer-request-handler.ts
+++ b/apps/cloudfront-functions/src/viewer-request-handler.ts
@@ -15,9 +15,15 @@ const handler = (
 
     // Check if the uri refers to a gitbook assets
     const isGitbookAssets = uri.startsWith('/gitbook/docs');
+    const isWoff2 = uri.endsWith('.woff2'); // woff2 is a font format
 
     // Add the .html extension if missing
-    if (!uri.endsWith('/') && !isGitbookAssets && !/\.[a-zA-Z]+$/.test(uri)) {
+    if (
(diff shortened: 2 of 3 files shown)


apps/cloudfront-functions/src/viewer-request-handler.ts
```diff
@@ -15,9 +15,15 @@ const handler = (
...
     const isGitbookAssets = uri.startsWith('/gitbook/docs');
+    const isWoff2 = uri.endsWith('.woff2'); // woff2 is a font format
 
     // Add the .html extension if missing
-    if (!uri.endsWith('/') && !isGitbookAssets && !/\.[a-zA-Z]+$/.test(uri)) {
+    if (
+      !uri.endsWith('/') &&
+      !isGitbookAssets &&
+      !isWoff2 &&
+      !/\.[a-zA-Z]+$/.test(uri)
```
datalek, line 25: I would generalize it as follows:

```suggestion
      !/\.[a-zA-Z0-9]+$/.test(uri)
```

apps/cloudfront-functions/src/viewer-request-handler.ts
```diff
@@ -15,9 +15,15 @@ const handler = (
 
     // Check if the uri refers to a gitbook assets
     const isGitbookAssets = uri.startsWith('/gitbook/docs');
+    const isWoff2 = uri.endsWith('.woff2'); // woff2 is a font format
```
datalek, line 18: If you change the regex, this should no longer be needed

jeremygordillo: If I change the regex, this test case breaks:

```js
const example1 = '/i-s/g/mo/v1.0';
expect(handler(makeEvent(example1)).uri).toBe(`${example1}.html`);
```

datalek: Great, then it's better to specify the extension as you've already done

.changeset/eighty-toys-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"cloudfront-functions": patch
+---
+
+handle woff2 font request
```
datalek, line 5:
```suggestion
Handle woff2 font request
```

datalek (approved):

datalek (approved):

PR #507, opened 2023-12-12
[DEV-866] Redirect to the main guide if no version is specified from the path

tommaso1:
#### List of Changes

Now the guide URL that doesn't contain a version links to the guide marked as `main`.

#### Motivation and Context

Assuming that https://developer.pagopa.it/app-io/guides/io-guida-tecnica/v3.0 is the latest version of the technical guide, the visitor must be able to access that guide from the following URLs:

https://developer.pagopa.it/app-io/guides/io-guida-tecnica/v3.0

https://developer.pagopa.it/app-io/guides/io-guida-tecnica

For the latest version, you can omit the version in the path.

#### How Has This Been Tested?

go to `pago-pa/guides/sanp/` check now it shows the last guide instead of an error

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #560, opened 2024-01-11
[DEV-863]: rewrite URLs if they point to https://docs.pagopa.it

MikeAtUqido:
#### List of Changes

* Rewrite URLs if they point to https://docs.pagopa.it pages

#### Motivation and Context

We need to point only on our pages

- [x] Manually

#### Types of changes

#### Checklist:


packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -2,6 +2,15 @@ import { Config } from '@markdoc/markdoc';
...
 
+const getCorrectGuide = (guide: string) => {
+  switch (guide) {
+    case 'manuale-operativo-dei-servizi':
+      return 'manuale-servizi';
+    default:
+      return guide;
+  }
+};
+
```
MikeAtUqido, line 13: We need to verify if other guides' name changes between docs and our internal URLs

marcobottaro: @certevol is going to map space names from docs.pagopa.it to the developer portal

certevol: Added the link to the file in "DEV-863. "The main difference is the direct link on docs.pagopa to the version chosen as the main one. In some cases, the manual's name on docs.pagopa includes also the product name, which would be redundant in the developer portal's slug. Others can be aligned.

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -53,7 +82,42 @@ export class LinkAttr {
             spacePrefix.pathPrefix
           )
         : value;
-    } else return value;
+    } else if (value) {
```
MarcoPonchia, line 85: Maybe here can be process only `value` that include `DOCS_URL` or is necessary to parse all links ? 
Something like: 
```
} else if (value.includes(DOCS_URL)) {
```

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -53,7 +82,42 @@ export class LinkAttr {
...
+          .split('/')
+          .filter((p) => p !== '');
+        return finalUrl === `/${name}/${rest.join('/')}`;
+      });
+
+      if (guide) {
+        return guide.path;
+      }
+
+      return value;
```
MarcoPonchia, line 117: i think could be useful to add a test to cover this case

MarcoPonchia (changes requested):

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -2,6 +2,35 @@ import { Config } from '@markdoc/markdoc';
 import path from 'path';
 import { ParseContentConfig } from '../parseContent';
 
+const getCorrectGuide = (guide: string) => {
+  switch (guide) {
```
marcobottaro, line 6: I think that some of the cases mapped in the spreadsheet are missing

marcobottaro (changes requested):
Check the spreadsheet

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -2,6 +2,41 @@ import { Config } from '@markdoc/markdoc';
...
+    case 'gestionedeglierrori':
+      return 'errori';
+    case 'dizionario-dei-metadata':
+      return 'metadata';
+    case 'manuale-operativo-di-firma-con-io':
+      return 'manuale-operativo';
+    default:
+      return guide;
+  }
+};
```
datalek, line 38: Wouldn't it be better to define the mapping when creating a guide? They could be retrieved from the Config, which contains the information injected by the caller.

apps/nextjs-website/src/_contents/products.ts
```diff
@@ -99,4 +99,24 @@ export const spaceToPrefixMap = pipe(
...
+  'kit-di-comunicazione-per-gli-enti': 'kit-comunicazione',
+  'io-come-aderire': 'accordi-adesione',
+  'f.a.q.-per-integratori': 'knowledge-base',
+  'modello-di-integrazione-di-piattaforma-notifiche': 'modello-di-integrazione',
+  'manuale-operativo-back-office-pagopa-ente-creditore': 'manuale-bo-ec',
+  'manuale-bo-pagopa-psp': 'manuale-bo-psp',
+  gestionedeglierrori: 'errori',
+  'dizionario-dei-metadata': 'metadata',
+  'manuale-operativo-di-firma-con-io': 'manuale-operativo',
+};
```
datalek, line 120: I would simplify it like this:

```TypeScript
export const urlRewritesMap = {
  'https://docs.pagopa.it/manuale-operativo-dei-servizi': 'manuale-servizi',
  'https://docs.pagopa.it/kb-enti': 'supporto-agli-enti',
  'https://domain.com/i-can/put/here/any/path': 'and-is-going/to/be-replaced/with-this-one',
  ...
};
```

By doing this, the logic in `gitbook-docs` can be simplified using a straightforward replacement. For any given URL, you can simply replace the left side with the corresponding right side. 

# Example
Given the following url `https://docs.pagopa.it/modello-di-integrazione-di-piattaforma-notifiche/v/v2.0/` and the following rewrites map:
```TypeScript
export const urlRewritesMap = {
  'https://docs.pagopa.it/modello-di-integrazione-di-piattaforma-notifiche/v/': '/send/guides/modello-di-integrazione/`
  ...
};
```
The system will produce the following url `/send/guides/modello-di-integrazione/v2.0/`

packages/gitbook-docs/src/markdoc/attributes.ts
```diff
@@ -53,7 +58,34 @@ export class LinkAttr {
...
+      const finalUrl = cleanUrl.replace(currentGuide, correctGuide);
+
+      const guides = parseContentConfig.gitBookPagesWithTitle;
+
+      const guide = guides.find((g) => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const [_product, _path, name, _version, ...rest] = g.path
+          .split('/')
+          .filter((p) => p !== '');
+        return finalUrl === `/${name}/${rest.join('/')}`;
```
datalek, line 78: There are too many assumptions about how the URL is formatted; I would perform a simple replace based on the urlRewritesMap to apply the rewrites.

MarcoPonchia (approved):

marcobottaro (approved):

MikeAtUqido:
This PR depends on https://github.com/pagopa/developer-portal/pull/507

apps/nextjs-website/src/_contents/products.ts
```diff
@@ -99,4 +99,47 @@ export const spaceToPrefixMap = pipe(
   RA.uniq(Eq.struct({ spaceId: s.Eq, pathPrefix: s.Eq }))
 );
 
+/**
+ * Contains the mapping between the docs.pago.it url and the developer portal url.
```
kin0992, line 103:
```suggestion
 * Contains the mapping between the docs.pagopa.it url and the developer portal url.
```

kin0992 (changes requested):

datalek (approved):

kin0992 (approved):

PR #622, opened 2024-02-12
[DEV-1350] Add SonarCloud scan to CI workflow

kin0992:
#### List of Changes

- Add scan step to code review workflow

#### Motivation and Context

We want to improve the analysis on code, using external tools

#### Types of changes

#### Checklist:


sonar-project.properties
```diff
@@ -0,0 +1,3 @@
+sonar.projectKey=pagopa_developer-portal
+sonar.organization=pagopa
+sonar.sources=apps/**/src, packages/gitbook-docs/src
```
datalek, line 3: Why not as following?

```suggestion
sonar.sources=apps/**/src, packages/**/src
```

kin0992: I wasn't sure that running the scan on `eslint-config` package would give useful information.

PR #686, opened 2024-02-28
[DEV-1274] Add Update Static Content plugin to Strapi to trigger static site deployment

marcobottaro:
#### List of Changes

Add [Update Static Content](https://market.strapi.io/plugins/strapi-plugin-update-static-content) plugin to Strapi to trigger static site deployment.

#### Motivation and Context

The editor needs to trigger static site deployment to see the changes made with Strapi CMS

Manually (without triggering any deployment)

Configuration:
image

Deployments list:
image

#### Types of changes

#### Checklist:


apps/strapi-cms/.env.default
```diff
@@ -15,3 +15,10 @@ AWS_BUCKET_ENDPOINT=
 AWS_BUCKET_NAME=
 AWS_BUCKET_FULL_URL=
 CDN_URL=
+
+# UPDATE STATIC CONTENT PLUGIN
+USC_GITHUB_PAT=
+USC_REPO_OWNER=
+USC_REPO_NAME=
+USC_WORKFLOW_ID=
+USC_TARGET_BRANCH=
```
kin0992, line 24: Why do you think the `USC` prefix is valuable?

marcobottaro: For two reasons: first of all the prefix lets the user identify the constants related to the USC plugin, secondary for consistency with the constants of AWS plugin

kin0992: You are not convincing me 😄 I mean, there are variable that, IMHO, should be general, without the need to specify the plugin or whatever uses it. It's an environment variable, not a plugin variable; that's why I'd call them differently.
E.g. `REPO_OWNER` instead of `USC_REPO_OWNER`.

The name should be as clear as possible and describe what piece of information that variable bring to the application: e.g. `USC_WORKFLOW_ID` could be something like `DEPLOY_WEBSITE_WORKFLOW_ID`, which tells me that variable contains the id of the `Deploy Website` workflow.

datalek (approved):

PR #689, opened 2024-02-28
[DEV-1462] - Terraform create log group for CMS

maurodandrea:
Terraform create log group for CMS Strapi and use it in ECS Task definition

#### List of Changes
Terraform create log group for CMS Strapi and use it in ECS Task definition

#### Motivation and Context
Terraform create log group for CMS Strapi and use it in ECS Task definition

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [X] My change not requires a change to the documentation.


.infrastructure/iam.tf
```diff
@@ -148,6 +148,14 @@ data "aws_iam_policy_document" "ecs_task_execution" {
...
   }
 
+  statement {
+    effect = "Allow"
+    actions = [
+      "logs:CreateLogGroup"
+    ]
+    resources = ["*"]
+  }
+
```
datalek, line 158: Avoid to define the resource `*` unless strictly necessary, follow always the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

I suggest to use the [CloudWatch Log Group created by default](https://github.com/terraform-aws-modules/terraform-aws-ecs/blob/master/docs/README.md#container-definition:~:text=The%20default%20behavior%20of%20the%20container%20definition%20module%20is%20to%20create%20the%20CloudWatch%20log%20group%20on%20behalf%20of%20users%20to%20ensure%20it%20can%20be%20fully%20managed%20through%20Terraform) by the [service module](https://github.com/pagopa/developer-portal/blob/9217615/.infrastructure/ecs.tf#L52-L79) if the `container_definitions` is defined or using a Log Group created via IaC, instead of having ECS create them separately. This ensures that all resources are connected to the Infrastructure as Code (IaC). By doing so, we can define the retention policy, which I'm uncertain can be achieved if ECS creates the resources, attach the tag and so on.

datalek (changes requested):

.infrastructure/iam.tf
maurodandrea: as agreed, the "strapi-cms" log group will be created in Terraform and this log group will be used in the taks-def

.infrastructure/task-definitions/cms_app.json.tpl
```diff
@@ -10,8 +10,7 @@
 	"logConfiguration": {
         "logDriver": "awslogs",
         "options": {
-            "awslogs-create-group": "true",
-            "awslogs-group": "/ecs/cms-task-def",
+            "awslogs-group": "strapi-cms",
```
datalek, line 13: I would provide it as variable
```suggestion
            "awslogs-group": "${log_group}",
```

datalek (changes requested):

.infrastructure/task-definitions/cms_app.json.tpl
maurodandrea: > I would provide it as variable

Ok, I must insert the var "cms_log_group" with value "strapi-cms" in the "variables.tf" file and "ecs.tf" file

.infrastructure/ecs.tf
```diff
@@ -35,6 +35,7 @@ data "template_file" "cms_app" {
     workflow_id          = "deploy_website.yaml"
     target_branch        = "main"
     github_pat           = module.secret_cms_github_pat.ssm_parameter_arn
+    log_group            = "strapi-cms"
```
datalek, line 38:
```suggestion
    log_group            = modue.cms_log_group.cloudwatch_log_group_name
```

datalek (changes requested):

.infrastructure/ecs.tf
maurodandrea: update commit with module.cms_log_group.cloudwatch_log_group_name

datalek (approved):

PR #704, opened 2024-03-05
Webinar questions review

datalek:
#### List of Changes

- Update the structure of `WebinarQuestion`
- Update the structure of `WebinarQuestionUpdate`
- Add tests related to webinar question update

#### Motivation and Context

Increases the robustness of the solution and prevent type casting

Unit test and locally

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
+++ b/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
@@ -49,8 +49,7 @@ const WebinarQuestionsTemplate = ({
   else {
     const userName = `${user.attributes['given_name']} ${user.attributes['family_name']}`;
     const sortedQuestions = [...data].sort(
-      (a: WebinarQuestion, b: WebinarQuestion) =>
-        b.createdAt.getTime() - a.createdAt.getTime()
+      (a, b) => b.id.createdAt.getTime() - a.id.createdAt.getTime()
     );
 
     return (
@@ -74,25 +73,27 @@ const WebinarQuestionsTemplate = ({
               <TableBody>
                 {sortedQuestions.map((webinarQuestion) => (
                   <WebinarQuestionRow
-                    key={webinarQuestion.createdAt.toISOString()}
+                    key={webinarQuestion.id.createdAt.toISOString()}
                     question={webinarQuestion}
                     userName={userName}
                     onHide={async (hide) =>
                       await updateWebinarQuestion({
-                        webinarId: webinarQuestion.webinarId,
-                        createdAt: webinarQuestion.createdAt,
-                        hiddenBy: hide
-                          ? { operation: 'update', value: userName }
-                          : { operation: 'remove' },
+                        id: webinarQuestion.id,
+                        updates: {
+                          hiddenBy: hide
+                            ? { operation: 'update', value: userName }
+                            : { operation: 'remove' },
+                        },
                       })
                     }
                     onHighlight={async (highlight) =>
                       await updateWebinarQuestion({
-                        webinarId: webinarQuestion.webinarId,
-                        createdAt: webinarQuestion.createdAt,
-                        highlightedBy: highlight
-                          ? { operation: 'update', value: userName }
-                          : { operation: 'remove' },
+                        id: webinarQuestion.id,
+                        updates: {
+                          highlightedBy: highlight
+                            ? { operation: 'update', value: userName }
+                            : { operation: 'remove' },
+                        },
                       })
                     }
                   />
--- a/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
+++ b/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
@@ -13,7 +13,10 @@ import { Webinar } from '@/lib/types/webinar';
 import { WebinarQuestion } from '@/lib/webinars/webinarQuestions';
 import { useWebinar } from '@/helpers/webinar.helpers';
 import { useEffect } from 'react';
-import { getWebinarQuestionList, updateQuestion } from '@/lib/webinarApi';
+import {
+  getWebinarQuestionList,
+  updateWebinarQuestion,
+} from '@/lib/webinarApi';
 import { useTranslations } from 'next-intl';
 import Spinner from '@/components/atoms/Spinner/Spinner';
 import useSWR from 'swr';
@@ -69,24 +72,28 @@ const WebinarQuestionsTemplate = ({
                 </TableRow>
               </TableHead>
               <TableBody>
-                {sortedQuestions.map((row) => (
+                {sortedQuestions.map((webinarQuestion) => (
                   <WebinarQuestionRow
-                    key={row.createdAt.toJSON()}
-                    question={row}
+                    key={webinarQuestion.createdAt.toJSON()}
+                    question={webinarQuestion}
                     userName={userName}
                     onHide={async (hide) =>
-                      await updateQuestion(
-                        row,
-                        'hide',
-                        hide ? userName : undefined
-                      )
+                      await updateWebinarQuestion({
+                        webinarId: webinarQuestion.webinarId,
+                        createdAt: webinarQuestion.createdAt,
+                        hiddenBy: hide
+                          ? { operation: 'update', value: userName }
+                          : { operation: 'remove' },
+                      })
                     }
                     onHighlight={async (highlight) =>
-                      await updateQuestion(
-                        row,
-                        'highlight',
-                        highlight ? userName : undefined
-                      )
+                      await updateWebinarQuestion({
+                        webinarId: webinarQuestion.webinarId,
+                        createdAt: webinarQuestion.createdAt,
+                        highlightedBy: highlight
+                          ? { operation: 'update', value: userName }
+                          : { operation: 'remove' },
+                      })
                     }
                   />
                 ))}
--- a/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
+++ b/apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
@@ -74,7 +74,7 @@ const WebinarQuestionsTemplate = ({
               <TableBody>
                 {sortedQuestions.map((webinarQuestion) => (
                   <WebinarQuestionRow
-                    key={webinarQuestion.createdAt.toJSON()}
+                    key={webinarQuestion.createdAt.toISOString()}
                     question={webinarQuestion}
                     userName={userName}
                     onHide={async (hide) =>
--- a/apps/nextjs-website/src/lib/webinars/__tests__/webinarQuestions.test.ts
+++ b/apps/nextjs-website/src/lib/webinars/__tests__/webinarQuestions.test.ts
@@ -14,11 +14,15 @@ import {
 import { makeDynamodbItemFromWebinarQuestion } from '../dynamodb/codec';
 
 const aWebinarQuestion = {
-  webinarId: 'aWebinarId',
+  id: {
+    slug: 'aWebinarId',
+    createdAt: new Date(0),
+  },
   question: 'aQuestion',
-  createdAt: new Date(),
-  hiddenBy: 'aHiddenBy',
-  highlightedBy: 'aHighlightedBy',
+};
+const aInsertWebinarQuestion = {
+  slug: aWebinarQuestion.id.slug,
+  question: aWebinarQuestion.question,
 };
 const aDynamoDBItem = makeDynamodbItemFromWebinarQuestion({
   ...aWebinarQuestion,
@@ -48,7 +52,7 @@ describe('webinarQuestions', () => {
   describe('insertWebinarQuestion', () => {
     it('should send dynamodb put command', async () => {
       const { env, dynamoDBClientMock, nowDateMock } = makeTestWebinarEnv();
-      const actual = await insertWebinarQuestion(aWebinarQuestion)(env)();
+      const actual = await insertWebinarQuestion(aInsertWebinarQuestion)(env)();
       const expected = E.right(undefined);
 
       expect(dynamoDBClientMock.send).toBeCalledTimes(1);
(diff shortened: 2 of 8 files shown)


apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
```diff
@@ -69,24 +72,28 @@ const WebinarQuestionsTemplate = ({
                 </TableRow>
               </TableHead>
               <TableBody>
-                {sortedQuestions.map((row) => (
+                {sortedQuestions.map((webinarQuestion) => (
                   <WebinarQuestionRow
-                    key={row.createdAt.toJSON()}
-                    question={row}
+                    key={webinarQuestion.createdAt.toJSON()}
```
kin0992, line 77:
```suggestion
                    key={webinarQuestion.createdAt.toISOString()}
```
I think that function is clearer than the `toJSON`. And [the output is the same](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toJSON)

kin0992 (changes requested):

apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
```diff
@@ -69,24 +72,28 @@ const WebinarQuestionsTemplate = ({
...
                     onHide={async (hide) =>
-                      await updateQuestion(
-                        row,
-                        'hide',
-                        hide ? userName : undefined
-                      )
+                      await updateWebinarQuestion({
+                        webinarId: webinarQuestion.webinarId,
+                        createdAt: webinarQuestion.createdAt,
+                        hiddenBy: hide
```
tommaso1, line 84: I would change the name of the param, hiddenBy to me is the user making the operation, not the operation itself, I'd use something like webinarQuestionUpdate, with an additional param to trak if its hidden or highlighted

apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
```diff
@@ -69,24 +72,28 @@ const WebinarQuestionsTemplate = ({
...
                     onHighlight={async (highlight) =>
-                      await updateQuestion(
-                        row,
-                        'highlight',
-                        highlight ? userName : undefined
-                      )
+                      await updateWebinarQuestion({
+                        webinarId: webinarQuestion.webinarId,
+                        createdAt: webinarQuestion.createdAt,
+                        highlightedBy: highlight
```
tommaso1, line 93: Same as the comment before, the name it's not super clear given the object passed

apps/nextjs-website/src/lib/webinars/__tests__/webinarQuestions.test.ts
```diff
@@ -26,6 +31,7 @@ const makeTestWebinarEnv = () => {
   // default mock behaviour
   dynamoDBClientMock.send.mockImplementation((cmd) => {
     if (cmd instanceof PutItemCommand) return Promise.resolve({});
+    else if (cmd instanceof UpdateItemCommand) return Promise.resolve({});
     else if (cmd instanceof QueryCommand)
       return Promise.resolve({ Items: [aDynamoDBItem] });
     // eslint-disable-next-line functional/no-promise-reject
```
tommaso1, line 37: It's more concise and readable to use an async function instead of explicitly rejecting or resolving promises.

```suggestion
  dynamoDBClientMock.send.mockImplementation(async (cmd) => {
    if (cmd instanceof PutItemCommand) return {};
    else if (cmd instanceof UpdateItemCommand) return {};
    else if (cmd instanceof QueryCommand)
      return { Items: [aDynamoDBItem] };
    else throw new Error('Unsupported command')
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
     question: { S: input.question },
-    hiddenBy: input.hiddenBy ? { S: input.hiddenBy } : { NULL: true },
-    highlightedBy: input.highlightedBy
-      ? { S: input.highlightedBy }
-      : { NULL: true },
+    ...(input.hiddenBy && { hiddenBy: { S: input.hiddenBy } }),
+    ...(input.highlightedBy && { highlightedBy: { S: input.highlightedBy } }),
   });
+
+type UpdateExpressionItem<T> = {
```
tommaso1, line 63: Is it necessary to make this generic? It is used only in one place

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+  // The type of the accumulator of the reduce function
+  type Zero = {
+    // the set command; e.g.: set fieldName0 = :fieldName0
+    readonly set?: string;
+    // the remove command; e.g.: remove fieldName0
+    readonly remove?: string;
+    // if ExpressionAttributeValues is empty the system raises a runtime error
+    readonly ExpressionAttributeValues?: UpdateItemCommandInput['ExpressionAttributeValues'];
+  };
+  const zero: Zero = {};
```
tommaso1, line 84: i find zero name confusing, i would rather call it initialValue, like in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce. 

I'd also rename the type to InitialValueType, 

```suggestion
  type InitialValueType = {
    // the set command; e.g.: set fieldName0 = :fieldName0
    readonly set?: string;
    // the remove command; e.g.: remove fieldName0
    readonly remove?: string;
    // if ExpressionAttributeValues is empty the system raises a runtime error
    readonly ExpressionAttributeValues?: UpdateItemCommandInput['ExpressionAttributeValues'];
  };


  const initialValue: InitialValueType = {};
```

or you could directly use it without naming to reduce the cognitive load to understand the function 

```suggestion
  const initialValue: {
    // the set command; e.g.: set fieldName0 = :fieldName0
    readonly set?: string;
    // the remove command; e.g.: remove fieldName0
    readonly remove?: string;
    // if ExpressionAttributeValues is empty the system raises a runtime error
    readonly ExpressionAttributeValues?: UpdateItemCommandInput['ExpressionAttributeValues'];
  } = {};
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+  'UpdateExpression' | 'ExpressionAttributeValues'
+> => {
+  // The type of the accumulator of the reduce function
+  type Zero = {
+    // the set command; e.g.: set fieldName0 = :fieldName0
+    readonly set?: string;
+    // the remove command; e.g.: remove fieldName0
+    readonly remove?: string;
+    // if ExpressionAttributeValues is empty the system raises a runtime error
+    readonly ExpressionAttributeValues?: UpdateItemCommandInput['ExpressionAttributeValues'];
```
tommaso1, line 82: Why this is uppercased? It is not clear if it's a type or a value, i'd use it with expressionAttributeValues and then remap it at the end of the function when the command is constructed, something like:


```suggestion
    readonly expressionAttributeValues?: UpdateItemCommandInput['ExpressionAttributeValues'];
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+  // ExpressionAttributeValues attribute
+  const { set, remove, ExpressionAttributeValues } = expressionList.reduce(
+    (acc, curr) => {
+      // handle update operations
+      if (curr.expression?.operation === 'update') {
+        // if set is empty initialize the set command, otherwise append to the existing one
+        const prefix = acc.set ? `${acc.set},` : 'set';
+        // the syntax is: set fieldName0 = :fieldName0, fieldNameN = :fieldNameN
+        const set = `${prefix} ${curr.fieldName} = :${curr.fieldName}`;
+        const ExpressionAttributeValues = {
```
tommaso1, line 95:
```suggestion
        const expressionAttributeValues = {
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+          ...acc.ExpressionAttributeValues,
```
tommaso1, line 96:
```suggestion
          ...acc.expressionAttributeValues,
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+          [`:${curr.fieldName}`]: { S: curr.expression?.value },
+        };
+        return { ...acc, set, ExpressionAttributeValues };
```
tommaso1, line 99:
```suggestion
        return { ...acc, set, expressionAttributeValues };
```

apps/nextjs-website/src/lib/webinars/dynamodb/codec.ts
```diff
@@ -38,25 +47,89 @@ export const makeWebinarQuestionFromDynamodbItem = (
...
+        return { ...acc, remove };
+      }
+      // handle no operations
+      else return acc;
+    },
+    zero
+  );
+  return {
+    UpdateExpression: `${set ?? ''} ${remove ?? ''}`,
+    ExpressionAttributeValues,
```
tommaso1, line 116:
```suggestion
    ExpressionAttributeValues: expressionAttributeValues,
```

apps/nextjs-website/src/lib/webinars/webinarQuestions.ts
```diff
@@ -35,64 +33,50 @@ export type WebinarQuestion = {
...
+  WebinarQuestion,
+  'webinarId' | 'createdAt'
+> & {
+  readonly hiddenBy?: UpdateExpression<string>;
+  readonly highlightedBy?: UpdateExpression<string>;
+};
+
 export const insertWebinarQuestion = (question: InsertWebinarQuestion) =>
   pipe(
     R.ask<Pick<WebinarEnv, 'dynamoDBClient' | 'nowDate'>>(),
```
tommaso1, line 55: This might not be clear from people with not a so strong fp background that is reading values from WebinarEnv, i'd add a comment explaining this line to make it more readable

apps/nextjs-website/src/lib/webinars/webinarQuestions.ts
```diff
@@ -35,64 +33,50 @@ export type WebinarQuestion = {
...
-        },
+        Item: makeDynamodbItemFromWebinarQuestion({
+          webinarId: question.webinarId,
+          createdAt: createdAt,
+          question: question.question,
+        }),
       });
       return TE.tryCatch(() => dynamoDBClient.send(putCommand), E.toError);
     }),
     RTE.map(() => void 0)
```
tommaso1, line 69: I'd add a comment explainig the last line is used to discard the result of the computation, because we are interested only in its side effects.
BTW, I think that it would be less elegant but more ts style and readable to the average developer to just get the client and the date and sending the command

apps/nextjs-website/src/lib/webinars/webinarQuestions.ts
```diff
@@ -35,64 +33,50 @@ export type WebinarQuestion = {
...
   );
 
-export const updateWebinarQuestion = (
-  question: WebinarQuestion,
-  action: 'highlight' | 'hide',
-  by?: string
-) =>
+export const updateWebinarQuestion = (update: WebinarQuestionUpdate) =>
   pipe(
     R.ask<Pick<WebinarEnv, 'dynamoDBClient'>>(),
```
tommaso1, line 74: Same as before, I'd add a comment here

apps/nextjs-website/src/lib/webinars/webinarQuestions.ts
```diff
@@ -35,64 +33,50 @@ export type WebinarQuestion = {
...
-          })
-        : new UpdateItemCommand({
-            ...baseUpdateCommand,
-            UpdateExpression: `REMOVE #${field}`,
-          });
+        ...makeDynamodbUpdateFromWebinarQuestionUpdate(update),
+      });
       return TE.tryCatch(() => dynamoDBClient.send(updateCommand), E.toError);
     }),
     RTE.map(() => void 0)
```
tommaso1, line 82: And a comment here

apps/nextjs-website/src/components/organisms/WebinarQuestionsTemplate/WebinarQuestionsTemplate.tsx
datalek: Can you provide an example to help me understand your suggestion?

tommaso1: Sure! :)
I think that having a name `hiddenBy` and then an object like `{ operation: 'update', value: userName }` it's a little bit confusing. With a name like that I am expecting to find something related to the user who made the operation and not not the operation itself. 

So i thought we could have it named differently... something like `hiddenUpdateOperation`, or `higlighedUpdateOperation`

datalek: Oki. I revised the `WebinarQuestionUpdate` and `WebinarQuestion` types to clarify the usage of the `updateWebinarQuestion` function. Let me know if this clarification helps

marcobottaro (approved):
With this PR we are introducing a lot of complexity (as demonstrated by the fact that the body of the makeUpdateExpression method has almost a 1:1 ratio between lines of code and comments) and, consequently, decreasing the maintainability of the code base.
We certainly gain in terms of robustness, but I think that 130 lines of code to avoid an explicit cast is a lot :)

I approve the PR, but I think we should add a task to the backlog to simplify the code in the future.

tommaso1 (approved):

MikeAtUqido (approved):

PR #783, opened 2024-03-26
[DEV-1345] Create SNS topic and subscriptions for metric alarms

kin0992:
Create the SNS topic and the subscription used to send alerts to an email address.

#### List of Changes

- Create the SNS topic
- Create the SNS subscription
    - Add details of where the alerts should be delivered

#### Motivation and Context

We want to be notified in case of some alerts are activated.

#### Types of changes

#### Checklist:

### DIFF
--- a/apps/infrastructure/src/sns_topic.tf
+++ b/apps/infrastructure/src/sns_topic.tf
@@ -1,6 +1,6 @@
 resource "aws_sns_topic" "metric_alarm" {
   name         = "cloudwatch_metric_alarms"
-  display_name = "DevPortal Alarms Topic"
+  display_name = "DevPortal Alarms"
 }
 
 resource "aws_sns_topic_subscription" "metric_alarm" {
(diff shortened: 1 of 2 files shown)


apps/infrastructure/src/sns_topic.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_sns_topic" "metric_alarm" {
+  name         = "cloudwatch_metric_alarms"
+  display_name = "DevPortal Alarms Topic"
+}
+
+resource "aws_sns_topic_subscription" "metric_alarm" {
+  protocol  = "email"
+  endpoint  = format("devportal-alerts+%s@pagopa.it", var.environment)
```
kin0992, line 8: I've decided that sending email for both environment, using the email alias, could help us to identify possible issues also when we work on dev environment.

apps/infrastructure/src/sns_topic.tf
```diff
@@ -0,0 +1,10 @@
+resource "aws_sns_topic" "metric_alarm" {
+  name         = "cloudwatch_metric_alarms"
+  display_name = "DevPortal Alarms Topic"
```
kin0992, line 3:
```suggestion
  display_name = "DevPortal Alarms"
```

datalek (approved):
LGTM

PR #791, opened 2024-03-28
[DEV-1524] webinars backend

tommaso1:
#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #793, opened 2024-03-28
[DEV-1524]: render homepage webinars from Strapi

MikeAtUqido:
This PR depends on #791

#### List of Changes

Render webinars coming from Strapi API

#### Motivation and Context

The webinars displayed in the homepage were only statics

Manually

#### Types of changes

#### Checklist:


apps/nextjs-website/src/lib/homepage.ts
```diff
@@ -94,6 +97,15 @@ export const makeHomepageProps = (
       })
     ),
   },
+  webinars: strapiHomepage.data.attributes.webinars.data.map((webinar) => ({
+    ...webinar.attributes,
+    isVisibleInHome: true,
```
MarcoPonchia, line 102: I think this attribute should be removed from here and from `Webinar` type. I think is better to add  `isVisibleInHome` only in [`apps/nextjs-website/src/_contents/webinars.ts` ](https://github.com/pagopa/developer-portal/blob/9c447fd5406b394804f7883f5d2f02564bca31ce/apps/nextjs-website/src/_contents/webinars.ts#L41C36-L41C43) constant to manange static webinars like you did in line 118.
Something like this:
```
type StaticWebinar = Webinar & {isVisibleInHome: boolean}
export const webinars: readonly StaticWebinar[] = [
...
```

MarcoPonchia (changes requested):

apps/nextjs-website/src/app/page.tsx
```diff
@@ -30,13 +29,11 @@ const NotSsrWebinarsSection = dynamic(
...
 
 const Home = async () => {
-  const webinars = await getVisibleInHomeWebinars();
-
   const homepage = await getHomepageProps();
 
   return (
     <>
-      <NotSsrWebinarHeaderBanner webinars={webinars} />
+      <NotSsrWebinarHeaderBanner webinars={[...homepage.webinars]} />
```
MarcoPonchia, line 36: I cannot see the banner even if i properly set in Strapi a webinar in the future

apps/nextjs-website/src/lib/homepage.ts
```diff
@@ -49,6 +51,7 @@ export type HomepageProps = {
       readonly href: string;
     }[];
   };
+  readonly webinars: readonly Webinar[] | readonly StaticWebinar[];
```
MarcoPonchia, line 54: I think is better to use only  `Webinar` type because, `StaticWebinar` is needed only to manage transition to static from Strapi webinars and will be deleted in future.
```suggestion
  readonly webinars: readonly Webinar[];
```

MarcoPonchia (changes requested):

apps/nextjs-website/src/lib/homepage.ts
```diff
@@ -94,6 +97,14 @@ export const makeHomepageProps = (
       })
     ),
   },
+  webinars: strapiHomepage.data.attributes.webinars.data.map((webinar) => ({
```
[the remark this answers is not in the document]
MikeAtUqido: This is true only if the `FETCH_FROM_STRAPI` env variable is set to `true` in the other case the function called is `makeHomepagePropsFromStatic` where only `StaticWebinar`s are returned, isn't that correct?

apps/nextjs-website/src/lib/homepage.ts
MikeAtUqido: This is true only if the `FETCH_FROM_STRAPI` env variable is set to `true` in the other case the function called is `makeHomepagePropsFromStatic` where only `StaticWebinar`s are returned, isn't that correct?
I'd delete the `StaticWebinar` type only when the webinars will be fetched exclusively by Strapi, is that already happening?

MarcoPonchia (approved):

marcobottaro (approved):

apps/nextjs-website/src/lib/homepage.ts
MikeAtUqido: I removed the type as it's not needed

marcobottaro (approved):

PR #827, opened 2024-04-10
[DEV-1401] Use absolute urls for media saved on Strapi

jeremygordillo:
#### List of Changes

- Created custom provider `strapi-provider-upload-custom` following [offical guide](https://docs.strapi.io/dev-docs/providers#creating-providers)
- Added `strapi-provider-upload-custom` as Strapi dependency

> ⚠️ All previously uploaded images must be reuploaded with this provider enabled.

> ℹ️ Custom provider only override some logics of [official strapi local provider](https://github.com/strapi/strapi/blob/main/packages/providers/upload-local/src/index.ts)

#### Motivation and Context

In the local environment, all images uploaded to Strapi have relative URLs, and when we fetch data from the API, content with media fields is not displayed on the frontend app because it is hosted on a different domain.

To solve this, we need to set absolute URLs.

Manually. 

First, you need to set `UPLOAD_PLUGIN_PROVIDER=strapi-provider-upload-custom` on Strapi .env file and then run `npm run dev`.

Go to Strapi Admin panel and load any image from the Media Library or the Content Manager.
Use uploaded images as image of some content type.

Screenshot 2024-04-10 at 17 23 48

The following screenshot shows two slides with different image sources. The first src has an absolute URL while the second one doesn't because I haven't set it in Content Manager so the fallback is used.

Screenshot 2024-04-10 at 17 25 32

#### Types of changes

#### Checklist:


packages/strapi-provider-upload-custom/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+const prefixFileUrlWithBackendUrl = (fileURL?: string): string | undefined => {
+  const baseUrl = getAbsoluteServerUrl(strapi.config);
+  return !!fileURL && fileURL.startsWith('/')
+    ? `${baseUrl}${fileURL}`
+    : fileURL;
+};
+
+export const init: typeof ProviderLocal.init = (options) => {
+  const fns = ProviderLocal.init(options);
```
marcobottaro, line 12: What does `fns` stand for?

.changeset/long-hairs-boil.md
```diff
@@ -0,0 +1,6 @@
+---
+"strapi-cms": minor
+"strapi-provider-upload-custom": minor
```
marcobottaro, line 3: If you are creating a new package, shouldn't be a `major` release?

packages/strapi-provider-upload-custom/src/index.ts
jeremygordillo: Stands for `functions`

.changeset/long-hairs-boil.md
jeremygordillo: I only followed the [semver](https://semver.org/#summary) specs but i can change them if you prefer.

tommaso1 (approved):

.changeset/long-hairs-boil.md
kin0992: **suggestion**: since you are updating two packages, I'd create two different changesets just to have a meaningful changelog for both packages.

packages/strapi-provider-upload-custom/package.json
kin0992, line 1: @jeremygordillo please update CODEOWNERS file, adding yourself, and who you think should be owner of the `strapi-provider-upload-custom` package.

kin0992 (changes requested):

packages/strapi-provider-upload-custom/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  return {
+    ...fns,
+    async uploadStream(file) {
+      // eslint-disable-next-line functional/no-expression-statements
+      await fns.uploadStream(file);
+      // eslint-disable-next-line functional/no-expression-statements, functional/immutable-data, @typescript-eslint/no-non-null-assertion
```
datalek, line 19: **suggestion (non-blocking)**: If some rules doesn't make sense in this package you can override them ([here](https://github.com/pagopa/developer-portal/blob/main/packages/eslint-config/eslint-functional.js#L30-L34) an example)

MikeAtUqido (approved):

kin0992 (approved):
CODEOWNERS file LGTM

PR #855, opened 2024-05-13
[DEV-1617, DEV-1619]: move webinars subscriptions to DynamoDB

MikeAtUqido:
This PR depends on #856

#### List of Changes

* Move webinars subscriptions to Dynamo DB

#### Motivation and Context

We used to store the webinars subscriptions directly in the user attributes, but this object has a maximum length

- [x] Manually
- [x] Unit tests

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


apps/nextjs-website/src/lib/webinars/__tests__/webinarSubscriptions.test.ts
```diff
@@ -0,0 +1,115 @@
...
+      const actual = await insertWebinarSubscription(
+        aInsertWebinarSubscription.webinarId,
+        aInsertWebinarSubscription.username
+      )(env)();
+      const expected = E.right(undefined);
+
+      expect(dynamoDBClientMock.send).toBeCalledTimes(1);
+      expect(actual).toStrictEqual(expected);
+    });
+    it('should return error if send returns an error', async () => {
```
MarcoPonchia, line 58: Nice job to add a test to check error case here :+1:  But why this test is only here for adding a subscription and there are not similar test for delete subscription and get subscriptions cases ' :thinking:

apps/nextjs-website/src/helpers/user.helper.ts
```diff
@@ -11,17 +12,27 @@ export const useUser = () => {
...
+      const subscriptions = await getUserWebinarSubscriptions(
+        user.attributes.email
+      ).catch(() => []);
+
+      setLoading(false);
+      setAligned(true);
+      setUser({ ...user, webinarSubscriptions: subscriptions });
+    };
+
+    run();
```
MarcoPonchia, line 35: There is some particular reason to create `run` function that is only call here in `checkUser` instad of something like this:
```
  const checkUser = useCallback(async () => {
    const user = await Auth.currentAuthenticatedUser().catch(() => {
      setLoading(false);
      setUser(null);
      return null;
    });

    if (!user) {
      return;
    }

    const subscriptions = await getUserWebinarSubscriptions(
      user.attributes.email
    ).catch(() => []);

    setLoading(false);
    setAligned(true);
    setUser({ ...user, webinarSubscriptions: subscriptions });
  }, []);
```

MarcoPonchia:
There is an issue after the state of loading. As you can see in the video, both in while subscribing and unsubscribing, the loading phase is not managed properly, as you can see when i click on subscribe button correctly change state to loading but before change again in cancel subscription for a moment show subscribe button. (You can see it well using slow 3G).
[Screencast from 15-05-2024 10:25:19.webm](https://github.com/pagopa/developer-portal/assets/24475319/15fedf47-39b6-47ad-a004-a54779f4bba9)

MarcoPonchia (approved):

apps/nextjs-website/src/helpers/user.helper.ts
```diff
@@ -10,18 +11,24 @@ export const useUser = () => {
...
-        setLoading(false);
-        setAligned(true);
-        setUser(user);
-      })
-      .catch(() => {
-        setLoading(false);
-        setAligned(true);
-        setUser(null);
-      });
+  const checkUser = useCallback(async () => {
```
tommaso1, line 14: i would call this fetchUserAndSubscriptions
```suggestion
  const fetchUserAndSubscriptions = useCallback(async () => {
```

apps/nextjs-website/src/helpers/user.helper.ts
```diff
@@ -42,6 +49,11 @@ export const useUser = () => {
       });
   };
 
+  const reloadUser = useCallback(async () => {
+    setLoading(true);
+    await checkUser();
```
tommaso1, line 54:
```suggestion
    await fetchUserAndSubscriptions();
```

apps/nextjs-website/src/helpers/user.helper.ts
```diff
@@ -42,6 +49,11 @@ export const useUser = () => {
...
+  }, [checkUser]);
```
tommaso1, line 55:
```suggestion
  }, [fetchUserAndSubscriptions]);
```

tommaso1 (approved):
I'd rename fetchUserAndSubscriptions, but it can work also as it is, I'm therefore approving it anyway

apps/nextjs-website/src/lib/webinars/webinarSubscriptions.ts
```diff
@@ -0,0 +1,98 @@
...
+    }),
+    RTE.chainEitherK(({ Items }) =>
+      pipe(
+        // turn undefined to empty array
+        Items || [],
+        // decode the response
+        RA.map(WebinarSubscriptionDynamodbCodec.decode),
+        // turn Array<Either<_, _>> to Either<_, Array<_>>
+        RA.sequence(E.Applicative),
+        // map errors to error and dynamodb item to WebinarQuestion
```
marcobottaro, line 94:
```suggestion
        // map errors to error and dynamodb item to WebinarSubscription
```

apps/nextjs-website/src/lib/webinars/__tests__/webinarSubscriptions.test.ts
```diff
@@ -0,0 +1,156 @@
...
+      const { env } = makeTestWebinarEnv();
+      const actual = await listUserWebinarSubscriptions(
+        aWebinarSubscription.username
+      )(env)();
+      const expected = E.right([aWebinarSubscription]);
+
+      expect(actual).toStrictEqual(expected);
+    });
+
+    it('should return an empty list if questions are undefined', async () => {
```
marcobottaro, line 126:
```suggestion
    it('should return an empty list if subscriptions are undefined', async () => {
```

marcobottaro (changes requested):

marcobottaro:
The "Visualizza" button should be clickable

image

marcobottaro (approved):

PR #919, opened 2024-06-17
[DEV-1692] Create a common template for guides and solutions pages

MikeAtUqido:
#### List of Changes

* Create a common template for guides and solutions pages

#### Motivation and Context

The old guides page and the new solutions page are very similar and use the same components

Manually

#### Types of changes

#### Checklist:


apps/nextjs-website/src/components/templates/GitBookTemplate/GitBookTemplate.tsx
tommaso1, line 1: I think maybe it's better to split this component into 2, because like this you need to cast props multiple times inside the component adding a lot of complexity to it, having 2 components, one to render ProductGuidePageProps and one to render SolutionDetailPageProps would make it more readable even if it will duplicate some code

tommaso1 (changes requested):

apps/nextjs-website/src/helpers/breadcrumbs.helpers.ts
```diff
@@ -60,3 +60,21 @@ export function pageToBreadcrumbs(
     ...(paths || []),
   ];
 }
+
+export function solutionPageToBreadcrumbs(
```
MarcoPonchia, line 64: i think is not necessary to add a new function but can be used instead `pageToBreadcrumbs` above in this file

apps/nextjs-website/src/components/templates/GitBookTemplate/GitBookTemplate.tsx
MarcoPonchia: Both solutions and guide page should be rendered in the same way, so now that all of those cast in the initial commits are moved to the relative page i think that having this common template proper manage the shared render of the pages

apps/nextjs-website/src/helpers/breadcrumbs.helpers.ts
marcobottaro: I think it's a ghost function

apps/nextjs-website/src/helpers/breadcrumbs.helpers.ts
```diff
@@ -60,3 +60,21 @@ export function pageToBreadcrumbs(
...
+  paths?: readonly Path[]
+): readonly BreadCrumbSegment[] {
```
marcobottaro, line 66: In this case, breadcrumb is a single word, so the type should be renamed

marcobottaro (changes requested):

tommaso1 (approved):

marcobottaro (approved):

PR #908, opened 2024-06-10
Add solution page

MarcoPonchia:
depends on: #907  #903 

#### List of Changes

Add solutions codec
Add solutions fetch
Align Media type issue
Add solution page

#### Motivation and Context

User can see solution coming from strapi

#### How Has This Been Tested?

In order to try this PR must create a Solution in strapi then you can test navigation on: http://localhost:3000/solutions/{solutionSlug}

#### Screenshots (if appropriate):
screenshot-localhost_3000-2024 06 10-10_46_17

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #932, opened 2024-06-26
[DEV-1674] Solution detail page

MikeAtUqido:
This PR depends on #908, #919

#### List of Changes

Add solution details page

#### Motivation and Context

The application was missing a page to display the solution

Manually

PagoPA-DevPortal-Soluzione-1

#### Types of changes

#### Checklist:


apps/nextjs-website/src/_contents/makeDocs.ts
```diff
@@ -107,3 +113,19 @@ export const makeGuide = ({
     parseDocOrThrow
   );
 };
+
+export const makeSolution = ({ solution, bannerLinks }: SolutionDefinition) => {
+  const solutionPath = `/solutions/${solution.slug}`;
```
MarcoPonchia, line 118: As you can see in video i got an error while navigating inside solutions detail, the problem is that is mounted a wrong html. 
[Screencast from 27-06-2024 10:37:27.webm](https://github.com/pagopa/developer-portal/assets/24475319/02641dca-de05-4c1c-9133-158b94edfd50)

One of the issue i think is due to path prefix.
```suggestion
  const solutionPath = `/solutions/${solution.slug}/details`;
```
But is not enough to fix the problem. As in guide and tutorial route should be different! I think solution detail page path should be something like `apps/nextjs-website/src/app/solutions/[solutionSlug]/details/[...solutionDetailPage]/page.tsx` and maybe some more fixes are needed to manage redirect and fragment properly :thinking:

MarcoPonchia (changes requested):

MikeAtUqido:
Closing in favor of #933

PR #974, opened 2024-07-09
[CAI-98] - Chatbot Vector Index on s3

mdciri:
#### List of Changes

- Added saving and loading of the vector index on AWS S3 bucket
- Updated chatbot params and prompts
- Updated chatbot script 

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.


apps/chatbot/src/modules/vector_database.py
```diff
@@ -134,20 +144,30 @@ def build_automerging_index(
...
         service_context=merging_context
     )
-    automerging_index.storage_context.persist(
-        persist_dir=save_dir
-    )
+    if s3_bucket_name:
+        automerging_index.storage_context.persist(
+            persist_dir=f"{s3_bucket_name}/{save_dir}",
+            fs = s3fs.S3FileSystem(
+                endpoint_url="https://s3.eu-west-3.amazonaws.com",
```
christian-calabrese, line 151: If you agree, I would get the region from the config file or an env variable

apps/chatbot/src/modules/vector_database.py
```diff
@@ -162,12 +182,26 @@ def load_automerging_index(
...
-        ),
-        service_context=merging_context,
-    )
+    if s3_bucket_name:
+
+        automerging_index = load_index_from_storage(
+            StorageContext.from_defaults(
+                persist_dir = f"{s3_bucket_name}/{save_dir}",
+                fs = s3fs.S3FileSystem(
+                    endpoint_url="https://s3.eu-west-3.amazonaws.com",
```
christian-calabrese, line 191: Same as above

christian-calabrese (approved):
LGTM

PR #1014, opened 2024-07-23
[DEV-1829]  Hide stats in solution if not present

MarBert:
#### List of Changes

Hide stats in solution if not present

#### Motivation and Context

There is an empty area if there are no stats

On localhost

BEFORE
image

AFTER
image

#### Types of changes

#### Checklist:

### DIFF
--- a/.changeset/wet-rocks-prove.md
+++ b/.changeset/wet-rocks-prove.md
@@ -2,4 +2,4 @@
 "nextjs-website": patch
 ---
 
-Change check for stats in solution
+Hide the entire statistics section on the Solution page when there are no values ​​to show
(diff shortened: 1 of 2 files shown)


.changeset/wet-rocks-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"nextjs-website": patch
+---
+
+Change check for stats in solution
```
marcobottaro, line 5: This is how you solved the problem, I think you should say something like
```suggestion
Hide the entire statistics section on the Solution page when there are no values ​​to show
```

marcobottaro (changes requested):

marcobottaro (approved):

PR #1057, opened 2024-08-22
[CAI-95] - Fix chatbot bedrock deployment

christian-calabrese:
#### List of Changes

Providers used to deploy the chatbot module have been set to strictly deploy resources to Paris region

#### Motivation and Context

The resources managed by the awscc provider were being created in Ireland

#### Types of changes

#### Checklist:


apps/infrastructure/src/modules/chatbot/lambda_chatbot.tf
```diff
@@ -18,7 +18,7 @@ module "lambda_function" {
   package_type  = "Image"
   architectures = ["x86_64"]
 
-  image_uri = module.ecr.repository_url
+  image_uri = "${module.ecr.repository_url}:latest"
```
batdevis, line 21: Maybe we couldn't use "latest" tag, because we can't use any cached version of the image.

christian-calabrese: As discussed on slack, we'll stick to the usage of `latest` tag until a more thoughtful way is implemented

batdevis (approved):

PR #1097, opened 2024-09-06
[CAI-100] Chatbot/dynamodb infrastructure

batdevis:
#### List of Changes
Chatbot/dynamodb Terraform infrastructure

#### Motivation and Context
Chatbot needs a dynamodb table in order to save questions and answers

### DIFF
--- a/apps/infrastructure/src/modules/chatbot/dynamodb.tf
+++ b/apps/infrastructure/src/modules/chatbot/dynamodb.tf
@@ -1,0 +1,24 @@
+# Chatbot queries
+
+module "dynamodb_chatbot_queries" {
+  source = "git::github.com/terraform-aws-modules/terraform-aws-dynamodb-table.git?ref=715399dbe24f6443820bf5de80f6100b35d56355" # v4.0.0
+
+  billing_mode                = "PAY_PER_REQUEST"
+  deletion_protection_enabled = true
+
+  name                           = "ChatbotQueries"
+  hash_key                       = "id"
+  range_key                      = "createdAt"
+  server_side_encryption_enabled = true
+
+  attributes = [
+    {
+      name = "id"
+      type = "S"
+    },
+    {
+      name = "createdAt"
+      type = "S"
+    },
+  ]
+}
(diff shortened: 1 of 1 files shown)


apps/infrastructure/src/modules/chatbot/dynamodb.tf
```diff
@@ -0,0 +1,28 @@
...
+
+module "dynamodb_chatbot_queries" {
+  source = "git::github.com/terraform-aws-modules/terraform-aws-dynamodb-table.git?ref=715399dbe24f6443820bf5de80f6100b35d56355" # v4.0.0
+
+  billing_mode                = "PAY_PER_REQUEST"
+  deletion_protection_enabled = true
+
+  name                           = "ChatbotQueries"
+  hash_key                       = "id"
+  range_key                      = "createdAt"
```
christian-calabrese, line 11: What is the id going to contain? Is it a unique id of a message or the session id?

If it's not already clear, I think it may be beneficiary to write a very brief RFC about the data model and how the table is going to be queried. 

After that it can be easier to choose the index (or indices) to accomodate the biggest number of direct queries against full scans

apps/infrastructure/src/modules/chatbot/dynamodb.tf
```diff
@@ -0,0 +1,28 @@
...
+      type = "S"
+    },
+    {
+      name = "createdAt"
+      type = "S"
+    },
+    {
+      name = "data"
+      type = "S"
+    },
```
christian-calabrese, line 26: The terraform resource forces to only define attributes that make part of the primary index (hash + sort key) or global/local secondary indices ([source](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#dynamodb-table-attributes))

At this point, I was thinking whether it would be feasible to extract the content of the data map into single fields.
This way, we can select the ones that will be actually indexed and include them in this list of attributes afterwards

This, of course, after having decided what kind of queries and indices the application needs (see previous comment)

apps/infrastructure/src/modules/chatbot/dynamodb.tf
batdevis: it's a unique id of a message.

christian-calabrese (approved):
LGTM as temporary dynamodb table to test queries

PR #1183, opened 2024-10-07
[CAI-118] Presidio

mdciri:
#### List of Changes

Add presidio to chatbot module to mask the Personally identifiable information (PII) entities

#### Motivation and Context

In this way, we can store a user's conversation masking all the PII in it for privacy reasons.

Jupyter notebook

#### Types of changes

#### Checklist:


batdevis (approved):

apps/chatbot/src/modules/models.py
```diff
@@ -22,7 +22,7 @@
 assert PROVIDER in ["aws", "google"]
 
 
-GOOGLE_PARAM_NAME = os.getenv("CHB_GOOGLE_API_KEY")
+GOOGLE_PARAM_NAME = os.getenv("GOOGLE_PARAM_NAME")
```
christian-calabrese, line 25: The environment variable has been previously named CHB_GOOGLE_API_KEY to be compliant with the standard we used in the others
```suggestion
GOOGLE_PARAM_NAME = os.getenv("CHB_GOOGLE_API_KEY")
```

apps/chatbot/src/modules/presidio.py
```diff
@@ -0,0 +1,215 @@
...
+            self.provider = NlpEngineProvider(conf_file=self.config)
+        elif isinstance(self.config, dict):
+            self.provider = NlpEngineProvider(nlp_configuration=self.config)
+        else:
+            raise ValueError("Error! config should be a path or a dictionary.")
+        nlp_engine = self.provider.create_engine()
+        self.nlp_engine = nlp_engine
+        self.analyzer = AnalyzerEngine(
+            nlp_engine = self.nlp_engine,
+            supported_languages = ["it", "en", "es", "fr", "de"],
```
christian-calabrese, line 122: Is it actually necessary to support other languages than italian?

apps/chatbot/src/modules/presidio.py
```diff
@@ -0,0 +1,215 @@
...
+    def class_name(cls) -> str:
+        return "PresidioPIINodePostprocessor"
+
+
+    def detect_language(self, text: str) -> str:
+
+        try:
+            lang_list = detect_langs(text)
+            for i in range(len(lang_list)-1, -1, -1):
+                if lang_list[i].lang not in ["it", "en", "es", "fr", "de"]:
```
christian-calabrese, line 140: ditto

apps/chatbot/src/modules/presidio.py
mdciri: I do not know... There isn't anymore a block language, so a user can write using any language. Presidio masks PII entites only if the language is one of its inputs (see `detect_pii` method). So, I assumed that a normal PagoPA user could speak one of the main european languages.

apps/chatbot/src/modules/presidio.py
mdciri: see previous comment

apps/chatbot/src/modules/models.py
mdciri: ok. Anyway, we need a better way to share better the env variables

batdevis: we have the [.env.example](https://github.com/pagopa/developer-portal/blob/main/apps/chatbot/.env.example) file

christian-calabrese (approved):
LGTM

apps/chatbot/src/modules/utils.py
```diff
@@ -8,7 +8,7 @@
 
 AWS_ACCESS_KEY_ID = os.getenv("CHB_AWS_ACCESS_KEY_ID")
 AWS_SECRET_ACCESS_KEY = os.getenv("CHB_AWS_SECRET_ACCESS_KEY")
-AWS_DEFAUL_REGION = os.getenv("GOOGLE_AND_REDIS_AWS_DEFAULT_REGION")
+AWS_DEFAUL_REGION = os.getenv("CHB_GOOGLE_AWS_DEFAULT_REGION")
```
batdevis, line 11: ssm parameters are in the aws default region

```suggestion
AWS_DEFAULT_REGION = os.getenv("CHB_AWS_DEFAULT_REGION")
```

apps/chatbot/src/modules/utils.py
```diff
@@ -8,7 +8,7 @@
...
 
 
 def get_ssm_parameter(name: str, default: str | None = None) -> str | None:
```
batdevis, line 14: change `AWS_DEFAUL_REGION` with `AWS_DEFAULT_REGION` in row 26

apps/chatbot/src/modules/vector_database.py
```diff
@@ -52,6 +52,7 @@
     aredis.ConnectionPool.from_url(REDIS_URL)
 )
 REDIS_INDEX_NAME = os.getenv("CHB_REDIS_INDEX_NAME")
+INDEX_ID = os.getenv("CHB_LLAMAINDEX_INDEX_ID")
```
christian-calabrese, line 55: To avoid modifying the entire set of environment variables during automatic indexing, I would use an ssm parameter.
This way, the CHB_LLAMAINDEX_INDEX_ID would contain the name of the ssm parameter. 

NB: it needs the import of the get_ssm_parameter method
```suggestion
INDEX_ID = get_ssm_parameter(os.getenv("CHB_LLAMAINDEX_INDEX_ID"))
```

apps/chatbot/.env.example
```diff
@@ -5,11 +5,16 @@ LOG_LEVEL=DEBUG
 CHB_AWS_ACCESS_KEY_ID=...
 CHB_AWS_SECRET_ACCESS_KEY=...
 CHB_AWS_DEFAULT_REGION=eu-west-3
+CHB_GOOGLE_AWS_DEFAULT_REGION=eu-south-1
+CHB_REDIS_AWS_DEFAULT_REGION=eu-south-1
```
christian-calabrese, line 9: Why do we need a separate variable for GOOGLE and REDIS regions?

Shall we use for example CHB_AWS_DEFAULT_REGION=eu-south-1 and CHB_BEDROCK_AWS_REGION=eu-west-3?

apps/chatbot/src/modules/utils.py
mdciri: no

apps/chatbot/.env.example
mdciri: ok, I can do this change

apps/chatbot/src/modules/vector_database.py
mdciri: ok

apps/chatbot/src/modules/utils.py
christian-calabrese: I think @batdevis meant that there's a typo in the name: DEFAU**L**_REGION instead of DEFAU**LT**_REGION

batdevis: yes

batdevis (approved):

christian-calabrese (approved):
LGTM

christian-calabrese (approved):

PR #1193, opened 2024-10-10
[CAI-185] Chatbot/docker compose with Redis and DynamoDB for local development

batdevis:
#### List of Changes
* docker compose with Redis and DynamoDB for local development
* fix to create_vector feature

#### Motivation and Context
The team needs to test features and bugfix locally first.

Locally.

#### Types of changes

#### Checklist:

- [ ] add auto creation of dynamodb tables


apps/chatbot/src/modules/vector_database.py
```diff
@@ -212,8 +229,8 @@ def create_documentation(
     # with open("full_text.txt", "w") as f:
     #     f.write(full_text)
 
-    logging.info(f"Number of documents with content: {len(documents)}")
-    logging.info(f"Number of empty pages in the documentation: {len(empty_pages)}. These are left out.")
+    logging.info(f"[vector_database.py] Number of documents with content: {len(documents)}")
```
christian-calabrese, line 232: I think we can use the log formatter to print the name of the script and much more
more info [here](https://stackoverflow.com/questions/533048/how-to-log-source-file-name-and-line-number-in-python)

apps/infrastructure/src/modules/chatbot/dynamodb.tf
```diff
@@ -41,5 +41,19 @@ module "dynamodb_chatbot_sessions" {
...
+
+  # GSI for query on created_at
+  global_secondary_index {
+    name            = "SessionsByCreatedAtIndex"
+    hash_key        = "userId"
+    range_key       = "createdAt"
+    read_capacity   = 5
+    write_capacity  = 5
+    projection_type = "ALL"
+  }
```
christian-calabrese:
```suggestion
  global_secondary_indexes = [{
    name            = "SessionsByCreatedAtIndex"
    hash_key        = "userId"
    range_key       = "createdAt"
    read_capacity   = 5
    write_capacity  = 5
    projection_type = "ALL"
  }]
```

apps/infrastructure/src/modules/chatbot/dynamodb.tf
```diff
@@ -41,5 +41,19 @@ module "dynamodb_chatbot_sessions" {
...
+    },
   ]
+
+  # GSI for query on created_at
+  global_secondary_index {
+    name            = "SessionsByCreatedAtIndex"
+    hash_key        = "userId"
+    range_key       = "createdAt"
+    read_capacity   = 5
+    write_capacity  = 5
```
christian-calabrese: Should be removed since we are not using the PROVISIONED billing_mode
```suggestion
```

apps/infrastructure/src/modules/chatbot/dynamodb.tf
```diff
@@ -41,5 +41,19 @@ module "dynamodb_chatbot_sessions" {
...
       type = "S"
     },
+    {
+      name = "createdAt"
+      type = "S"
+    },
   ]
+
+  # GSI for query on created_at
+  global_secondary_index {
```
christian-calabrese: I think we can use a local_secondary_index cause the hash key is the same. This way we also reduce cost

apps/chatbot/docker/files/.aws/credentials
```diff
@@ -0,0 +1,3 @@
+[default]
+aws_access_key_id = 123
+aws_secret_access_key = xyz
```
christian-calabrese, line 3: If the docker container needs to call actual AWS resources (for example bedrock), I think this may result in unauthorized access.
I've commonly seen using the option C from [this answer](https://stackoverflow.com/a/56077990). This way, the container uses your AWS credentials and configuration

apps/chatbot/.env.example
```diff
@@ -21,3 +21,6 @@ CHB_ENGINE_SIMILARITY_CUTOFF=...
 CHB_ENGINE_USE_ASYNC=...
 CHB_ENGINE_USE_STREAMING=...
 CHB_QUERY_TABLE_PREFIX=chatbot-local
+GOOGLE_AND_REDIS_AWS_DEFAULT_REGION=eu-west-3
```
christian-calabrese, line 24: I'm not sure about the naming:
We currently have a region we use for everything (eu-south-1) and a region we use only for bedrock and S3 (eu-west-3)

What do you think about changing the names to CHB_AWS_DEFAULT_REGION and CHB_AWS_BEDROCK_REGION? cc @mdciri

apps/chatbot/src/modules/vector_database.py
batdevis: You are right
(but now I remove the logs)

apps/chatbot/.env.example
batdevis: same for me, I think @mdciri will update it in another pr

apps/chatbot/docker/files/.aws/credentials
batdevis: I tested and it works because our code uses `CHB_AWS_*` .
The problem is the local dynamodb that needs the dummy profile.

apps/infrastructure/src/modules/chatbot/dynamodb.tf
batdevis: done

apps/chatbot/docker/files/.aws/credentials
christian-calabrese: Doesn't it work with real credentials?

christian-calabrese (approved):

christian-calabrese (approved):

marcobottaro (approved):

apps/chatbot/docker/compose.yaml
```diff
@@ -0,0 +1,55 @@
...
+      - ntw
+
+  create_index:
+    build:
+      context: ..
+      dockerfile: docker/app.local.Dockerfile
+    ports:
+      - "8080:8080"
+    volumes:
+      - ..:/app
```
marcobottaro, line 45:
```suggestion
      - ..:/app
      - ./build-devp/out:/app/build-devp/out
```

marcobottaro (changes requested):

mdciri (approved):

marcobottaro (approved):

christian-calabrese (approved):

PR #1195, opened 2024-10-11
[CAI-182] fix(chatbot): api gateway permits DELETE requests

batdevis:
#### List of Changes
added DELETE to allowed methods list in api gateway

#### Motivation and Context
We have a DELETE API.


christian-calabrese (approved):
LGTM

PR #1198, opened 2024-10-11
[CAI-183] Chatbot Patch query api

batdevis:
#### List of Changes
Update `badAnswer` query field.

curl example:
```
curl -X 'PATCH' "http://localhost:8080/sessions/$SESSION_ID/queries/$ID" \
    -H 'Content-Type: application/json'   -d '{
    "badAnswer": true
  }'
``` 

#### Motivation and Context
We want to save the feedback about the answer.


christian-calabrese (approved):

apps/chatbot/src/app/main.py
```diff
@@ -115,12 +119,12 @@ def find_or_create_session(userId: str, now: datetime.datetime):
   if userId is None:
     userId = '-'
   
-  SESSION_MAX_DURATION_DAYS = int(os.getenv('SESSION_MAX_DURATION_DAYS', '1'))
+  SESSION_MAX_DURATION_DAYS = float(os.getenv('CHB_SESSION_MAX_DURATION_DAYS', '1'))
```
mdciri, line 122: why `float`?

mdciri (approved):

PR #1209, opened 2024-10-18
[DEV-1977] Add environment variable to disable custom validators during Strapi's data transfer

marcobottaro:
#### List of Changes

Add environment variable in .env.default
Add an exit guard to custom validators

#### Motivation and Context

The transfer command fails if validators are enabled (it's a Strapi bug https://github.com/strapi/strapi/issues/19383)

Manually

#### Types of changes

#### Checklist:


apps/strapi-cms/.env.default
```diff
@@ -5,6 +5,8 @@ API_TOKEN_SALT=tobemodified
 ADMIN_JWT_SECRET=tobemodified
 TRANSFER_TOKEN_SALT=tobemodified
 JWT_SECRET=tobemodified
+# Set this variable to True to disable custom validators during data transfer
```
tommaso1, line 8:
```suggestion
# Set this variable to True to disable custom validators during data transfer. Allowed values: True, False
```

tommaso1 (changes requested):

tommaso1 (approved):

PR #1212, opened 2024-10-25
[CAI-222] - Mistral

mdciri:
#### List of Changes

- Use Mistral as main LLM from AWS Bedrock in Paris region
- index id with date 
- delete old index when creating a new one

#### Motivation and Context

- LLM for privacy reasons
- index id for automatic re-indexing

jupyter notebook and gradio app

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation.


apps/chatbot/src/modules/vector_database.py
```diff
@@ -41,16 +42,23 @@
...
+EMBEDDING_DIMS = {
+    "models/text-embedding-004": 768,
+    "cohere.embed-multilingual-v3": 1024
+}
 REDIS_SCHEMA = IndexSchema.from_dict({
-    "index": {"name": REDIS_INDEX_NAME, "prefix": "index/vector"},
+    "index": {"name": f"{REDIS_INDEX_NAME}_{INDEX_ID}", "prefix": f"index_{INDEX_ID}/vector"},
     "fields": [
         {"name": "id", "type": "tag", "attrs": {"sortable": False}},
         {"name": "doc_id", "type": "tag", "attrs": {"sortable": False}},
```
christian-calabrese, line 64: Is this actually needed anymore?

batdevis (approved):

apps/chatbot/src/modules/vector_database.py
mdciri: it is for the the current schema. otherwise it would not load. the plan is to change it after a new re-indexing

christian-calabrese (approved):

.gitmodules
batdevis, line 1: I think you have not to remove the submodule

batdevis (approved):

PR #1213, opened 2024-10-29
[CAI-162] Chatbot/Save masked questions and answers using Presidio

batdevis:
#### List of Changes
Save masked questions and answers using Presidio.

#### Motivation and Context
Privacy.

my local environment

### DIFF
--- a/apps/chatbot/src/modules/chatbot.py
+++ b/apps/chatbot/src/modules/chatbot.py
@@ -127,10 +127,13 @@ class Chatbot():
     
 
     def mask_pii(self, message: str) -> str:
-        try:
-            return self.pii.mask_pii(message)
-        except Exception as e:
-            logging.warning(f"[chatbot.py - mask_pii] exception in mask_pii: {e}")
+        if USE_PRESIDIO:
+            try:
+                return self.pii.mask_pii(message)
+            except Exception as e:
+                logging.warning(f"[chatbot.py - mask_pii] exception in mask_pii: {e}")
+        else:
+            return message
 
 
     def generate(self, query_str: str) -> str:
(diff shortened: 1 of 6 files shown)


apps/chatbot/src/modules/chatbot.py
```diff
@@ -127,10 +127,13 @@ def _unmask_reference(self, response_str: str, nodes) -> str:
     
 
     def mask_pii(self, message: str) -> str:
-        try:
-            return self.pii.mask_pii(message)
-        except Exception as e:
-            logging.warning(f"[chatbot.py - mask_pii] exception in mask_pii: {e}")
+        if USE_PRESIDIO:
```
mdciri, line 130: is this env var still necessary or can we remove it?

apps/chatbot/src/modules/vector_database.py
```diff
@@ -235,7 +235,7 @@ def build_automerging_index_redis(
     logging.info(f"[vector_database.py - build_automerging_index_redis] hash_table_{INDEX_ID} is now on Redis.")
 
     logging.info(f"[vector_database.py - build_automerging_index_redis] Creating index {NEW_INDEX_ID} ...")
-    nodes = Settings.node_parser.get_nodes_from_documents(documents)
+    nodes = Settings.node_parser.get_nodes_from_documents(documents[:2])
```
mdciri, line 238: remove [:2]

mdciri (changes requested):

apps/chatbot/src/modules/chatbot.py
batdevis: for now we keep it

apps/chatbot/src/modules/vector_database.py
batdevis: done

mdciri (approved):

PR #1318, opened 2025-01-21
[DEV-2122] implement in-memory caching for CMS API data retrieval

tommaso1:
#### List of Changes

- Introduced a new `cache.ts` file to manage an in-memory cache for API responses.
- Updated multiple functions in `cmsApi.ts` to utilize the caching mechanism, improving performance by reducing redundant API calls.
- Each API function now caches results for 15 minutes (900 seconds) using a unique cache key generated by `getCacheKey`.
- Enhanced type safety by replacing `any` with `unknown` in the cache implementation.

This change optimizes data fetching and reduces load on the CMS API.

#### Types of changes

#### Checklist:


apps/nextjs-website/src/lib/cmsApi.ts
```diff
@@ -90,46 +133,92 @@ let areUrlReplaceMapCached = false;
 export const getCachedUrlReplaceMapProps = async () => {
   if (!areUrlReplaceMapCached) {
     // eslint-disable-next-line functional/no-expression-statements
-    cachedUrlReplaceMapProps = await getUrlReplaceMapProps();
+    cachedUrlReplaceMapProps = await withCache(
+      getCacheKey('getUrlReplaceMapProps'),
+      async () => getUrlReplaceMapProps(),
+      900
+    );
```
MarcoPonchia, line 140: Wach out we are double caching this data because `getCachedUrlReplaceMapProps` use `getUrlReplaceMapProps` that are already get cached so i think is better to remove `getCachedUrlReplaceMapProps` and refactor all components where is present using `getUrlReplaceMapProps`

MarcoPonchia (changes requested):

apps/nextjs-website/src/lib/cmsApi.ts
```diff
@@ -48,88 +49,157 @@ const buildEnv = pipe(
...
 export const getHomepageProps = async () => {
-  const strapiHomepage = await fetchHomepage(buildEnv);
-  return makeHomepageProps(strapiHomepage);
+  return withCache(
+    getCacheKey('getHomepageProps'),
+    async () => {
+      const strapiHomepage = await fetchHomepage(buildEnv);
+      return makeHomepageProps(strapiHomepage);
+    },
+    900 // 15 minutes in seconds
```
MarcoPonchia, line 58: maybe we should convert this into a env var ot at  least a constant in this file
```suggestion
    900 // 15 minutes in seconds
```

MarcoPonchia (changes requested):

apps/nextjs-website/src/lib/cache.ts
```diff
@@ -0,0 +1,41 @@
+// In-memory cache store
+interface CacheEntry<T> {
+  readonly data: T;
+  readonly expiry: number;
+}
+
+// Replace any with unknown for better type safety
```
marcobottaro, line 7:
```suggestion
```

.changeset/tall-cats-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"nextjs-website": patch
+---
+
+Caching strapi apis
```
marcobottaro, line 5:
```suggestion
Implement caching for API responses
```

apps/nextjs-website/src/lib/cache.ts
```diff
@@ -0,0 +1,41 @@
...
+const cacheStore = new Map<string, CacheEntry<unknown>>();
+
+export const getCacheKey = (prefix: string, ...args: readonly string[]) =>
+  `cache:${prefix}:${args.join(':')}`;
+
+export const withCache = async <T>(
+  key: string,
+  fetchFn: () => Promise<T>,
+  ttl: number = parseInt(process.env.REDIS_CACHE_TTL || '300')
```
marcobottaro, line 16: What's ttl unit of measurement? I think you should add the suffix `_in_seconds` to the function parameter, add the suffix `_IN_SECONDS` to the environment variable name and add it to `.env.default`.

marcobottaro: With the standalone version of Next.js will we need the `NEXT_PUBLIC_` prefix?
Can we consider it the `_DEFAULT_VALUE`?

apps/nextjs-website/src/lib/cmsApi.ts
```diff
@@ -49,84 +50,149 @@ const buildEnv = pipe(
   })
 );
 
+const CACHE_EXPIRY_SECONDS = 900;  // 15 minutes in seconds
```
marcobottaro, line 53: Do we really need this constant here? What's the difference with `process.env.REDIS_CACHE_TTL`?
Is there a reason why you set 900 seconds here and 300 seconds as default value in `apps/nextjs-website/src/lib/cache.ts`?

marcobottaro (changes requested):

marcobottaro (approved):

PR #1345, opened 2025-02-12
[DEV-2135] fix images in tables

MarBert:
#### List of Changes

Change the doc parsing to show images in <td> tags
#### Motivation and Context

Images in table are seen as text

On localhost  (images are broken)
image

#### Types of changes

#### Checklist:

### DIFF
--- a/packages/gitbook-docs/src/__tests__/parseContent.test.ts
+++ b/packages/gitbook-docs/src/__tests__/parseContent.test.ts
@@ -869,7 +869,7 @@ describe('parseContent', () => {
     ]);
   });
 
-  it('should parse html table with image', () => {
+  it('should parse html table with an image inside a table data cell', () => {
     const table =
       '<table data-header-hidden>' +
       '<thead><tr>' +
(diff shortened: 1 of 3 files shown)


packages/gitbook-docs/src/__tests__/parseContent.test.ts
```diff
@@ -869,6 +869,43 @@ describe('parseContent', () => {
     ]);
   });
 
+  it('should parse html table with image', () => {
```
marcobottaro, line 872:
```suggestion
  it('should parse html table with an image inside a table data cell', () => {
```

packages/gitbook-docs/src/parseContent.ts
```diff
@@ -44,7 +44,7 @@ const pairedHtmlTag = (tag: string) => ({
   replace: `{% ${tag}$1 %}$2{% /${tag} %}`,
 });
 const unpairedHtmlTag = (tag: string) => ({
-  regex: new RegExp(`<${tag}(.*?)>`, 'g'),
+  regex: new RegExp(`(?<!<td>)<${tag}(.*?)>`, 'g'),
```
marcobottaro, line 47: You should also consider a table data cell with some attributes inside like `<td class="bold">` and add it to the previous test

marcobottaro (changes requested):
The changeset file is missing

PR #1371, opened 2025-03-05
Add uqido consultants as CODEOWNERS

christian-calabrese:
Add Uqido consultants as CODEOWNERS to make them autonomous on PR merges


marcobottaro:
@christian-calabrese I closed the redundant PR #1373 and I think you should add `[DEV-2194]` to the title of this PR

kin0992 (approved):
LGTM

I left just a **suggestion(non-blocking)**: we could clean up this file

We could remove every entry associated to the default owners. For any file or directory not explicitly specified in the CODEOWNERS file, the default owners will be the default reviewers

PR #1404, opened 2025-04-07
[DEV-2157] sync docs repo

tommaso1:
#### List of Changes

Add new pipeline to sync docs github repository into the dedicated s3 bucket. This allows the other pipelines to have fresh data to calculate sitemap with

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/devportal-docs-sync.yaml
+++ b/.github/workflows/devportal-docs-sync.yaml
@@ -3,7 +3,7 @@ name: DevPortal Docs Sync
 on:
   push:
     branches:
-      - DEV-2157
+      - poc-standalone
   workflow_dispatch:
 
 # Permissions needed for AWS OIDC authentication


.github/workflows/devportal-docs-sync.yaml
```diff
@@ -0,0 +1,39 @@
+name: DevPortal Docs Sync
+
+on:
+  push:
+    branches:
+      - DEV-2157
```
MarcoPonchia, line 6: I think this branch was set for test the PR but maybe is better to add comment to remember to update branches to `poc-standalone`.
This workflow will also work for multiple branch, like when will be used from both `main` and `uat` :thinking: ?

MarcoPonchia (changes requested):

MarcoPonchia (approved):

PR #1414, opened 2025-04-11
[CAI-389] AWS SSM vars prefix

batdevis:
#### List of Changes
insert CHB_AWS_SSM prefix for the env vars whose value is an SSM path

#### Motivation and Context
Simplify the process to understand the purpose of every environment variable.

In my local environment with the frontend and launching the test suite

#### Types of changes

#### Checklist:


apps/chatbot/.env.example
```diff
@@ -11,19 +11,19 @@ CHB_AWS_BEDROCK_EMBED_REGION=eu-central-1
 CHB_AWS_BEDROCK_LLM_REGION=eu-west-3
 CHB_AWS_DEFAULT_REGION=eu-south-1
 CHB_AWS_SECRET_ACCESS_KEY=...
-CHB_DOCUMENTATION_DIR=...
```
mdciri, line 14: why did you remove this one?

apps/chatbot/.env.example
```diff
@@ -56,6 +56,6 @@ POSTGRES_PASSWORD=postgres
 POSTGRES_PORT=5432
 POSTGRES_USER=postgres
 DATABASE_URL=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}:${POSTGRES_PORT}/${POSTGRES_DB}
-PYTHONPATH=$PWD
+PYTHONPATH=/app
```
mdciri, line 59: `/app` is correct for the docker container, `$PWD` otherwise. I think `$PWD` is correct for the docker container too, no?

mdciri (approved):

apps/chatbot/.env.example
batdevis: not removed but moved for alphabetically order

apps/chatbot/.env.example
batdevis: I'm not so sure. So we use `/app`, since the docker container is our main method for local development.

marcobottaro (approved):

PR #1451, opened 2025-05-26
[DEV-2296] refactor api section component

Sebastiano-Bertolin:
#### List of Changes
Refactor ApiSection to ApiRestSection 
Add a generic ApiSection

#### Motivation and Context
Have a generic ApiSection responsible to chose the right component to render depending on the API type.
As for now there is only ApiRestSection so the ApiSection returns by default this component.

Manually
The API page behaviour should be the same as the one in the parent branch

#### Types of changes

#### Checklist:


marcobottaro:
I fixed the prefix of the PR title `[DEV-2296]`

marcobottaro (approved):

PR #1463, opened 2025-06-06
[DEV-2310] turbo upgrade

tommaso1:
#### List of Changes

Upgrades turbo to the latest version to allow the upgrade of nextjs

#### Types of changes

#### Checklist:


apps/strapi-cms/package.json
```diff
@@ -35,8 +35,7 @@
   "devDependencies": {
     "eslint": "^8.50.0",
     "eslint-config-custom": "*",
-    "typescript": "^5.7.3",
-    "turbo": "^1.10.12"
```
marcobottaro, line 39: I think that this row was due because of the `turbo prune` command

apps/strapi-cms/.env.default
marcobottaro, line 1: Why did you removed this file?

package.json
```diff
@@ -19,6 +19,8 @@
   "devDependencies": {
     "@changesets/cli": "^2.27.1",
     "husky": "^8.0.3",
-    "turbo": "^1.10.12"
-  }
+    "turbo": "^2.5.3"
+  },
+  "packageManager": "npm@10.8.2",
```
marcobottaro, line 24: Why do we need to set an npm version here?

If this is necessary, I think you should also change `.node-version` file content and set `22.8.0` according to nodejs documentation: https://nodejs.org/dist/index.json

Does current Next and Strapi versions are compatible with npm `10.8.2`?

marcobottaro (changes requested):

apps/strapi-cms/package.json
kin0992: Turbo shouldn't be a workspace dependency, but in a mono repo, we use turbo to run some commands in any workspace.

so, the scripts of a single workspace (like the `strapi-cms` app) should not depend on turbo. 
you could accomplish what you are doing now, making explicit the dependency between tasks (in the `turbo.json` file and updating (if necessary) your GitHub workflows.

If you do so, you will boost performance because you will also use the turbo cache (if the `turbo.json` file is properly configured).

package.json
kin0992: i left a **note** about the `packageManager`: it is a good practice to add it to the `package.json` file.
This adds consistency and makes sure the CI uses the same version used during development.

Moreover, I guess this has been introduced running the codemod to migrate turbo.

PR #1589, opened 2025-07-30
[CAI-473] Add trigger to start reindex on sync ends

MarBert:
#### List of Changes

Adds a trigger to start the reindex workflow when the sync workflow ends

#### Types of changes

#### Checklist:

### DIFF
--- a/.github/workflows/chatbot_reindex.yaml
+++ b/.github/workflows/chatbot_reindex.yaml
@@ -14,6 +14,11 @@ on:
           - uat
           - prod
 
+  workflow_run:
+    workflows:
+      - Sync GitBook Docs
+    types: [completed]
+
   # Allows external webhook trigger
   repository_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/chatbot_reindex.yaml
```diff
@@ -14,6 +14,11 @@ on:
           - uat
           - prod
 
+  workflow_run:
+    workflows:
+      - Sync GitBook Docs
```
marcobottaro, line 19: Are you sure about this? As far as I know, in GitHub Actions, when using `workflow_run`, you should refer to the workflow by its file name (e.g., `sync_gitbook_docs.yaml`), not by its display name (e.g., `Sync GitBook Docs`). I'm not sure that using the name will work as expected.

marcobottaro (changes requested):

PR #1605, opened 2025-08-07
[CAI-532] Fix urls with #

MarBert:
#### List of Changes

Fix parsing for urls with #

on localhost
#### Types of changes

#### Checklist:


packages/gitbook-docs/src/helpers/documentationParsing.helper.ts
```diff
@@ -61,8 +61,7 @@ export function replaceUrl(
     });
   const lastPart = splitValue.at(-1) || '';
   const secondToLastPart = splitValue.at(-2) || '';
-  const name = lastPart.replace('.md', '');
-
+  const name = lastPart.replace('.md', '').split('#')[0];
```
marcobottaro, line 64: I think you should add a test using a url with an anchor

marcobottaro (changes requested):

PR #1682, opened 2025-09-12
[CAI 538] fix evaluation queue

batdevis:
#### List of Changes
Delete sqs message after elaboration

#### Motivation and Context
Sometimes evaluation is called double.

We will test it in development environment.

#### Types of changes

#### Checklist:

### DIFF
--- a/.changeset/deep-towns-sneeze.md
+++ b/.changeset/deep-towns-sneeze.md
@@ -1,0 +1,5 @@
+---
+"chatbot": patch
+---
+
+Fix repetitive score bug in the Langfuse GUI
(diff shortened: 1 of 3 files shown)


uolter:
@batdevis in theory it shouldn't be required to delete the messages. It is one of the advantages offered by lambda.

According to the official documentation [here](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#sqs-polling-behavior)
> With Amazon SQS event source mappings, Lambda polls the queue and invokes your function [synchronously](https://docs.aws.amazon.com/lambda/latest/dg/invocation-sync.html) with an event. Each event can contain a batch of multiple messages from the queue. Lambda receives these events one batch at a time, and invokes your function once for each batch. When your function successfully processes a batch, **Lambda deletes its messages from the queue**.

Do we have evidence that some messages are not deleted? May be they duplicated in the queue and my grouping within the trace-it is not working or not correct.

image

PR #2228, opened 2026-04-09
[DEV-3885] Removed strapi 5 moule

uolter:
#### Types of changes

#### Checklist:


README.md
marcobottaro, line 1: Why are you changing the this file?

marcobottaro (changes requested):

apps/infrastructure/src/main.tf
```diff
@@ -297,7 +297,7 @@ module "video_streaming" {
 module "strapi_v5" {
   source = "./modules/strapi5"
 
-  count = var.environment == "dev" ? 1 : 0
+  count = 0
```
marcobottaro: Is there a reason why you keep `strapi_v5` module with `count: 0` instead of removing it?

uolter: Is it not needed in UAT? I might set count = 1 when the environment is uat.

PR #2251, opened 2026-04-24
[DEV-4054] Fix login redirect 

MarBert:
#### List of Changes

This pull request fixes the redirect logic for users who are already logged in on the Next.js website. It ensures that users with a logged-in cookie are properly redirected and that the cookie is reset as needed.

Authentication and redirect fixes:

* Updated the middleware in `apps/nextjs-website/src/middleware.ts` to check if the `loggedInCookieName` cookie value is `'true'` before redirecting users away from guest paths. Now, when a logged-in user accesses a guest path, the system redirects them to the appropriate locale and resets the cookie by setting its value to `'false'` and expiring it immediately.
* Imported `baseUrl` from the config to correctly set the cookie domain when resetting the cookie.

Changelog:

* Added a changeset `.changeset/sunny-badgers-hug.md` describing the patch fix for the redirect when the logged-in cookie is set.

Localhost by setting the cookie while not logged in 
#### Types of changes

#### Checklist:


Sebastiano-Bertolin (approved):

marcobottaro (approved):

MarcoPonchia:
Testing functionally this PR i notice that sometimes Login not work until page is fully reload, otherwise issue remain i don't know if is correct. I add a video

https://github.com/user-attachments/assets/81ae5f31-1068-423d-a925-a6b733cdd58f

PR #2277, opened 2026-06-09
[DEV-4274] Add trigger for crowdin sync workflow in devportal-docs repository

Sebastiano-Bertolin:
This pull request adds automation to trigger the Crowdin source upload workflow in the docs repository when syncing Italian documentation to S3 in the production environment. It introduces a new input for a GitHub PAT token, updates the relevant GitHub Action, and ensures the token is passed from workflow secrets.

Key changes:

**Automation of Crowdin workflow:**
* Added a step in the `sync-gitbook-docs` GitHub Action to trigger the `upload_sources_to_crowdin.yml` workflow in the `pagopa/devportal-docs` repository when syncing Italian docs on production. This uses the `gh` CLI and is conditional on environment and locale.

**Action and workflow inputs:**
* Introduced a new optional `pat_token` input in `.github/actions/sync-gitbook-docs/action.yaml` to supply a GitHub PAT for authentication when triggering the Crowdin workflow.
* Updated the main and scheduled sync jobs in `.github/workflows/sync_gitbook_docs.yaml` to pass the `PAT_TOKEN` secret as the `pat_token` input to the action, ensuring secure authentication. [[1]](diffhunk://#diff-a5935f106fdf02ceeb3c9b808a95c0b9980b915d03724ae0ebbd300e5e30ce28R110) [[2]](diffhunk://#diff-a5935f106fdf02ceeb3c9b808a95c0b9980b915d03724ae0ebbd300e5e30ce28R153)


.github/workflows/sync_gitbook_docs.yaml
```diff
@@ -56,6 +56,11 @@ on:
         type: boolean
         required: true
         default: true
+      documentation_branch:
+        description: 'The branch of the documentation repository to sync (optional, default: main)'
+        type: string
+        required: false
+        default: 'main'
```
marcobottaro: I'm not sure about this input, we should discuss about it
anyway the default value shouldn't be `main`

```suggestion
        description: 'The branch of the documentation repository to sync (optional, default: docs/from-gitbook)'
        type: string
        required: false
        default: 'docs/from-gitbook'
```

.github/workflows/sync_gitbook_docs.yaml
```diff
@@ -190,3 +195,20 @@ jobs:
...
+  trigger_crowdin_sync_workflow:
+    name: Trigger Crowdin source upload in target repository
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Trigger Crowdin source upload workflow
+        env:
+          GH_TOKEN: ${{ secrets.PAT_TOKEN }}
+          PATHS_TO_UPLOAD: ${{ inputs.dir_names_filter }}
+          BRANCH: ${{ inputs.documentation_branch }}
```
marcobottaro: If you add this step in a separate job, it will run at the end of each workflow run, whether triggered manually or scheduled.
In the latter case, the inpus.* values will not be available

marcobottaro (changes requested):

.github/workflows/sync_gitbook_docs.yaml
Sebastiano-Bertolin: fixed

MarBert (approved):

MarBert (approved):

.github/workflows/sync_gitbook_docs.yaml
```diff
@@ -56,6 +56,11 @@ on:
         type: boolean
         required: true
         default: true
+      documentation_branch:
+        description: 'The branch of the documentation repository to sync (optional, default: docs/from-gitbook)'
+        type: string
+        required: false
+        default: 'docs/from-gitbook'
```
MarcoPonchia: I think this newly added input is never used in this workflow. 🤔 Maybe also the actions that require this param should be updated as well; otherwise, this input can be removed.

MarcoPonchia (changes requested):

MarcoPonchia (approved):

marcobottaro (approved):

